PR #10, opened 2022-03-13
Properly update HTTP port value

### LINKED ISSUE #9: [bug] cannot update HTTP port
- Changing the value of `Insecure HTTP Server Port` seems to actually change the HTTPS port
Screen Shot 2022-03-12 at 19 15 46
Screen Shot 2022-03-12 at 19 16 33

- Note I've kept the HTTPS port default, but modified HTTP to `27125`, while the dev console shows HTTP still at `27123` and _HTTPS_ at `27125`. Changing the `Secure HTTPS Server Port` value also modifies the actual HTTPS port.

- macOS 12.2.1
- Obsidian v0.13.33

ericzinnikas:
- Both HTTPS/HTTP inputs were updating the same (HTTPS) value
- Fixes #9

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -236,7 +236,7 @@ class LocalRestApiSettingTab extends PluginSettingTab {
       .addText((cb) =>
         cb
           .onChange((value) => {
-            this.plugin.settings.port = parseInt(value, 10);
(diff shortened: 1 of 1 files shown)


coddingtonbear:
Thanks for the fix, @ericzinnikas! :heart:

PR #44, opened 2022-11-29
feat: auto create dirs

### LINKED ISSUE #42: [Feature Request][`/vault/{filename}` PUT/POST] Create folder when not exist
Current: 500 Internal Server Error
Expect: Create the folder and file.

SettingDust:
fix #42

### DIFF
--- a/src/requestHandler.ts
+++ b/src/requestHandler.ts
@@ -68,10 +68,7 @@ export default class RequestHandler {
 
   requestIsAuthenticated(req: express.Request): boolean {
     const authorizationHeader = req.get("Authorization");
-    if (authorizationHeader === `Bearer ${this.settings.apiKey}`) {
-      return true;
-    }
-    return false;
+    return authorizationHeader === `Bearer ${this.settings.apiKey}`;
   }
 
   async authenticationMiddleware(
@@ -259,7 +256,7 @@ export default class RequestHandler {
       return;
     }
 
-    await createDirNotExist(filepath);
+    createDirNotExist(filepath);
     await this.app.vault.adapter.write(filepath, req.body);
 
     this.returnCannedResponse(res, { statusCode: 204 });
@@ -376,7 +373,7 @@ export default class RequestHandler {
       return;
     }
 
-    await createDirNotExist(filepath)
+    createDirNotExist(filepath)
     await this.app.vault.adapter.write(filepath, req.body);
 
     let fileContents = "";
--- a/src/requestHandler.ts
+++ b/src/requestHandler.ts
@@ -1,6 +1,14 @@
-import {apiVersion, App, CachedMetadata, Command, PluginManifest, prepareSimpleSearch, TFile,} from "obsidian";
+import {
+  apiVersion,
+  App,
+  CachedMetadata,
+  Command,
+  PluginManifest,
+  prepareSimpleSearch,
+  TFile,
+} from "obsidian";
 import periodicNotes from "obsidian-daily-notes-interface";
-import {getAPI as getDataviewAPI} from "obsidian-dataview";
+import { getAPI as getDataviewAPI } from "obsidian-dataview";
 
 import express from "express";
 import http from "http";
@@ -23,8 +31,8 @@ import {
   SearchJsonResponseItem,
   SearchResponseItem,
 } from "./types";
-import {findHeadingBoundary} from "./utils";
-import {CERT_NAME, ContentTypes, ERROR_CODE_MESSAGES} from "./constants";
+import { findHeadingBoundary } from "./utils";
+import { CERT_NAME, ContentTypes, ERROR_CODE_MESSAGES } from "./constants";
 import path from "node:path";
 
 export default class RequestHandler {
@@ -96,7 +104,7 @@ export default class RequestHandler {
     const cache = this.app.metadataCache.getFileCache(file);
 
     // Gather frontmatter & strip out positioning information
-    const frontmatter = {...(cache.frontmatter ?? {})};
+    const frontmatter = { ...(cache.frontmatter ?? {}) };
     delete frontmatter.position; // This just adds noise
 
     // Gather both in-line tags (hash'd) & frontmatter tags; strip
@@ -119,10 +127,10 @@ export default class RequestHandler {
   }
 
   getResponseMessage({
-                       statusCode = 400,
-                       message,
-                       errorCode,
-                     }: ErrorResponseDescriptor): string {
+    statusCode = 400,
+    message,
+    errorCode,
+  }: ErrorResponseDescriptor): string {
     let errorMessages: string[] = [];
     if (errorCode) {
       errorMessages.push(ERROR_CODE_MESSAGES[errorCode]);
@@ -136,7 +144,7 @@ export default class RequestHandler {
     return errorMessages.join("\n");
   }
 
-  getStatusCode({statusCode, errorCode}: ErrorResponseDescriptor): number {
+  getStatusCode({ statusCode, errorCode }: ErrorResponseDescriptor): number {
     if (statusCode) {
       return statusCode;
     }
@@ -145,14 +153,14 @@ export default class RequestHandler {
 
   returnCannedResponse(
     res: express.Response,
-    {statusCode, message, errorCode}: ErrorResponseDescriptor
+    { statusCode, message, errorCode }: ErrorResponseDescriptor
   ): void {
     const response: CannedResponse = {
-      message: this.getResponseMessage({statusCode, message, errorCode}),
+      message: this.getResponseMessage({ statusCode, message, errorCode }),
       errorCode: errorCode ?? statusCode * 100,
     };
 
-    res.status(this.getStatusCode({statusCode, errorCode})).json(response);
+    res.status(this.getStatusCode({ statusCode, errorCode })).json(response);
   }
 
   root(req: express.Request, res: express.Response): void {
@@ -191,7 +199,7 @@ export default class RequestHandler {
       files.sort();
 
       if (files.length === 0) {
-        this.returnCannedResponse(res, {statusCode: 404});
+        this.returnCannedResponse(res, { statusCode: 404 });
         return;
       }
 
@@ -271,7 +279,7 @@ export default class RequestHandler {
       if (file instanceof TFile) await this.app.vault.modify(file, req.body)
     }
 
-    this.returnCannedResponse(res, {statusCode: 204});
+    this.returnCannedResponse(res, { statusCode: 204 });
     return;
   }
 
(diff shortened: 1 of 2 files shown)


src/requestHandler.ts
```diff
@@ -267,7 +268,9 @@ export default class RequestHandler {
       return;
     }
 
-    await this.app.vault.adapter.write(path, req.body);
+    const directory = path.relative(process.cwd(), path.dirname(filepath))
```
coddingtonbear: I'm not totally sure I understand what you're trying to accomplish by calculating the relative path between the process's current working directory (which could be anywhere on the filesystem -- even somewhere unrelated to the path to your vault) and the directory name of the file you're sending in.  Care to elaborate?

The reason I ask is that I believe `this.app.vault.adapter.exists(path.dirname(filepath))` would be sufficient, no?  I'm under the impression that the vault adapter already assumes that file paths passed-in are relative to your vault root.

src/requestHandler.ts
```diff
@@ -267,7 +268,9 @@ export default class RequestHandler {
...
+    if (!await this.app.vault.adapter.exists(directory)) await this.app.vault.adapter.mkdir(directory)
```
coddingtonbear: I haven't checked, but can you confirm that `this.app.vault.adapter.mkdir` will create multiple path segments if multiple directories do not exist?  I mean -- say you're creating a file at `/alpha/beta/delta.md` but neither the `alpha` nor `beta` directories exist; will it create both?

src/requestHandler.ts
```diff
@@ -383,8 +386,12 @@ export default class RequestHandler {
       return;
     }
 
+    const directory = path.relative(process.cwd(), path.dirname(filepath))
+    if (!await this.app.vault.adapter.exists(directory)) await this.app.vault.adapter.mkdir(directory)
+    await this.app.vault.adapter.write(filepath, req.body);
```
coddingtonbear: Since we're doing this twice, we should probably consider wrapping lines 389:391 into their own function that we can call from multiple locations.

src/requestHandler.ts
SettingDust: Yup
image

But there is a bug of obsidian. It's will create a dir that use another separator
image

src/requestHandler.ts
SettingDust: Seems that I facing a problem is the dirname will return a full path. But now it's not. I have no idea what is happened. But I removed that now

src/requestHandler.ts
SettingDust: I've tried `replace(/[\/\\]/g, path.sep)`. But obsidian still create the non-exist dir. And can't delete caused it's not exist. Have to restart

coddingtonbear: I _bet_ what's happening there is that, instead of it creating intermediate path segments, it's creating a file with the name you've specified, and that name just happens to include a path segment in its name. I wonder if we need to recursively build up the directory path instead?

coddingtonbear:
@SettingDust -- I'm at the moment assuming you've run out of time to work on this pull request; so I'm going to close this request due to inactivity.  If you did find the time to pick this up again, though, feel free to re-open or ask me to once you've worked through the directory creation issues discussed above.

src/utils.ts
```diff
@@ -42,7 +43,7 @@ export function findHeadingBoundary(
 }
 
 
-export async function createDirNotExist(filepath: string) {
-  const directory = path.dirname(filepath)
-  if (!await this.app.vault.adapter.exists(directory)) await this.app.vault.adapter.mkdir(directory)
```
coddingtonbear: @SettingDust -- I think we really should be using the Obsidian API instead of the `fs` module where possible, including here.  Although in most cases we'll be operating on an actual filesystem, it's totally possible that Obsidian will in some cases abstract away something that _can't_ be accessed via the `fs` module.

Because I'm not sure it's clear -- what I'm fairly sure was happening before is that the `adapter.mkdir` method **does not** know how to handle recursively creating multiple directories, and was instead creating a single directory with the path separator (`/`) in its name.

To work around this limitation, though, is really easy -- all you need to do is split `directory` on the path separator, and in your code, create each necessary path segment, one at a time.

coddingtonbear (changes requested):

src/utils.ts
SettingDust: I've tried. Obsidian can't change current directory to children. So, create `a` then `a/b` still not working

cdloh: Can you not just create the directory structure from the top down?

ie 

```
await app.vault.createFolder("TestTopDir")
await app.vault.createFolder("TestTopDir/Inner")
```

Then do as @coddingtonbear suggested and loop through until you get to the base.

SettingDust: Talked. Creating `a` then `a/b` won't work as expected. I've tried XD

SettingDust:
Due to https://discord.com/channels/686053708261228577/840286264964022302/1104023861449728060

Obsidian 1.2.7 should fix this issue. But I need modify something by the guideline. 
https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Prefer+the+Vault+API+over+the+Adapter+API

SettingDust:
Should be fine in Obsidian 1.2.7

juangamnik:
> Should be fine in Obsidian 1.2.7

@SettingDust: So is this a feature finished or is there still something open (and if yes, what needs to be done?).

SettingDust:
> @SettingDust: So is this a feature finished or is there still something open (and if yes, what needs to be done?).

Need review from @coddingtonbear

juangamnik:
> Need review from @coddingtonbear

One question: why do you use `fs.mkdir` instead of `Vault.createFolder(path:string)` of the obsidian API?

SettingDust:
> One question: why do you use `fs.mkdir` instead of `Vault.createFolder(path:string)` of the obsidian API?

Sorry. I can't understand what are you saying. You can read reviews and comments for following current context

juangamnik:
> Sorry. I can't understand what are you saying. You can read reviews and comments for following current context

I resembled to the following statement of yours (above):

> Sry. I forgot this PR. I've fixed this issue by using fs.mkdir. Have pushed

But I see that in your [commit from last week](https://github.com/coddingtonbear/obsidian-local-rest-api/pull/44/commits/1e5f2e072451f671c907b1dc2768bfd7859b5c8b) that you already changed the implementation in `utils.ts` to use the obsidian API instead of direct file system access.

Looks good to me (but I didn't test it).

How can I test this fork in my obsidian installation?

SettingDust:
> How can I test this fork in my obsidian installation?

Clone the repo and build

src/requestHandler.ts
```diff
@@ -383,8 +373,11 @@ export default class RequestHandler {
       return;
     }
 
+    await createDirNotExist(filepath)
+    await this.app.vault.adapter.write(filepath, req.body);
```
coddingtonbear: I might be misreading this, but it kind of looks like were changing the implementation of `vaultPost` here from:

1. Checking if there's an existing file and reading its contents if necessary
2. Writing the file contents from (1) and then appending the incoming request body

to:

1. Creating the directory parent directories if they don't already exist.
2. Writing the request body to the file.
3. Checking if there's an existing file and reading its contents if necessary.
4. Writing the the file contents from (3) and appending the incoming request body.

I _think_ the right choice here is to just remove line 377 ((2) above), but I might be misunderstanding your intentions here.

coddingtonbear:
@juangamnik thanks for trying to help review this!  It's a little easier for both you and the person who you're reviewing the code for if you go to the "Files Changed" tab and add comments on the relevant lines.  Otherwise: it's a little difficult to piece together.  Thanks for having a look, though!

src/utils.ts
```diff
@@ -39,3 +40,12 @@ export function findHeadingBoundary(
 
   return null;
 }
+
+
+export async function createDirNotExist(filepath: string) {
+  try {
+    await this.app.vault.createFolder(path.dirname(filepath))
```
coddingtonbear: Reading through the comment you noted, @SettingDust (from https://discord.com/channels/686053708261228577/840286264964022302/1104023861449728060; copied here for clarity):

> You can run the code in  the developer console (Ctrl+Shift+I).
> 
> Testing in v1.2.7 shows that this.app.vault.adapter.mkdir can indeed create multiple path segments, even if the folder already exists:
> await this.app.vault.adapter.mkdir("a/b/c")
> 
> However, you should prefer the vault API (https://docs.obsidian.md/Plugins/Releasing/Plugin+guidelines#Prefer+the+Vault+API+over+the+Adapter+API), so you should use this.app.vault.createFolder. However, while this can also create multiple path segments, it does not like it if the folder already exists. So you could write:
> 
> ```
> if (!this.app.vault.getAbstractFileByPath("a/b/c")) {
>   await this.app.vault.createFolder("a/b/c")
> }
> ```
> 
> Alternatively, catch the exception thrown by it if the folder already exists (EAFP):
> 
> ```
> try {
>   await this.app.vault.createFolder("a/b/c")
> } catch {
>   // the folder already exists, but we don't care
> }
> ```

It sounds like this might fail if _any_ path segments already exist -- so if, for example, we were trying to write a file under `a/b/c/filename.md` and `a/b/` already existed, but `c/` did not, his phrasing in the comment you linked suggests that this *might* throw an error.  Can you verify that it doesn't?  If it does, it seems like that's something we should fix, and, if so, I think my earlier PR review comments might give you some tips.

SettingDust: Tried. I created `a` folder. Execute put `vault/a/b/c/d.md`. No error.

src/requestHandler.ts
SettingDust: You are right. MB XD

sbtourist:
Hello! What's the status of this PR? Is it ready to be merged and when will it be released? I'm also happy to help :)

SettingDust:
> Hello! What's the status of this PR? Is it ready to be merged and when will it be released? I'm also happy to help :)

Ready for merge

sbtourist:
@SettingDust Nice, thanks, hopefully @coddingtonbear can merge and release it soon.

coddingtonbear:
@sbtourist -- I think the only major thing preventing this from being merged/released is finding the time to install this locally and test it.  If you'd be willing to take that on, that would accellerate things.  Otherwise, it might be a couple weeks before I can find the time.

sbtourist:
@coddingtonbear Sure I can do that, should be quick.

SettingDust:
@sbtourist shouldn't relate to my changes. Difficult to say what's wrong with it. But it's fine in my local dev env

coddingtonbear:
@sbtourist -- I believe what you might need to do is run `npm i` before running `npm run dev`; without doing that, you'll be missing dependencies.

coddingtonbear:
These changes were merged (with some minor simplifications) as part of 3704f2c.  Thanks for your help with this, @SettingDust !

sbtourist:
No, I had run it. Unfortunately I've been taken away by unexpected stuff
and wasn't able to investigate more, but great to hear you managed to merge
it! Thanks!

PR #61, opened 2023-05-29
Allow auth with X-API-KEY header instead of Authorization header

derekantrican:
Resolves discussion #60

_I also have pending changes to turn this on & off via a setting. Let me know if you would like me to commit those._

_Part of me also wants to add a setting to make this user-customizable (eg something other than `X-API-KEY`) but maybe this is fine. Again, let me know your thoughts._

### DIFF
--- a/src/requestHandler.ts
+++ b/src/requestHandler.ts
@@ -79,6 +79,12 @@ export default class RequestHandler {
     if (authorizationHeader === `Bearer ${this.settings.apiKey}`) {
       return true;
     }
+    
+    const apiKeyHeader = req.get("X-API-KEY");
+    if (apiKeyHeader === this.settings.apiKey) {
+      return true;
+    }
+
     return false;
   }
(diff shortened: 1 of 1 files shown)


derekantrican:
I will imagine I will have to update some documentation somewhere - just point me in the right direction

derekantrican:
@coddingtonbear any opinions on this? Could this be merged?

coddingtonbear:
OK, I made similar changes to what you were using via https://github.com/coddingtonbear/obsidian-local-rest-api/pull/65 -- the main difference is that I went the approach of just making the name of the header containing the authorization information configurable.

This'll go out as part of the next release, which I hope to find some time to do within the next few days.

derekantrican:
Awesome! That was a lot faster than I expected!

I'll check it out when I next get a chance (either tomorrow or Monday) but sounds great

PR #77, opened 2023-10-12
Add support for type and recursive options

baileywickham:
Add support for directory and recursive parameters.
* `type=file | folder` filters for files/folders
* `recursive=true` returns the recursive list of files and directories

Let me know what else needs to be done to merge this. 

also, I had to change the obsidian npm import to `obsidian: latest` in the package.json to get things to work locally.

Thanks for the work on this!

### DIFF
--- a/src/requestHandler.ts
+++ b/src/requestHandler.ts
@@ -182,7 +182,7 @@ export default class RequestHandler {
   async _vaultGet(
     path: string,
     recursive: boolean,
-    directory: boolean,
+    type: "folder" | "file" | null,
     req: express.Request,
     res: express.Response
   ): Promise<void> {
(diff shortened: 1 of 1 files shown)


coddingtonbear:
Super curious what your use-case is here for this kind of thing?  It seems like if we're going to list items recursively, we're stepping away from the "REST-ful" nature of the API.  If this is something you need, maybe it might make sense for it to be handled via a separate endpoint?

baileywickham:
I'm good with moving it to a separate endpoint, did you have one in mind? I don't have a strong have a strong preference one way or another. 

For my use case: I am using your obsidian web plugin which I forked to add support for a "bookmarks" style directory. I want to have a dropdown of subdirectories of my Bookmarks directory in the "Create new note" menu so I can add the bookmark I want. Example: `Bookmarks/software`, `Bookmarks/security`, and I can select either directories from the chrome plugin. I'll upstream those changes if I can get them in good shape, they are sort of hacked on as of now

coddingtonbear:
I really appreciate you taking the time to put together a PR for this, and I'm sorry to say this, but I'm afraid I'm not convinced that this is something I want to support going forward, so I'm going to decline this pull request. 

The reasons that I've landed on this decision are that I really want to keep this project as close to being a standard REST API as possible so as to make it easier to maintain going-forward.  Little features like you're suggesting can totally help with niche use cases like yours, but they do make the maintenance overhead a little bit worse for each additional feature in the future.  Additionally, what you're trying to accomplish can be easily solved by just recursively dispatching REST requests for each subdirectory to list entries, so there _is_ a clear workaround that lines up a lot more closely with how REST APIs are intended to work.

Again, really sorry to not being up for maintaining this extension of the plugin's functionality for the future, but I hope you understand that complexity is the main problem with long-term maintenance on projects like this one.

Issue #88, opened 2023-12-19
bug: can’t write binary file 


rowthan:
i tried write a image file called by put method but faild.

i think it’s a problem in this method 
`_vaultPut` .  why it prevent write binary files?

rowthan:
i have made a pull request to solve this problem.  check it please #89

coddingtonbear:
I wonder if you could elaborate on the problem you might've been experiencing -- I'm vaguely aware that we do handle binary content _as long as the content-type_ header is set correctly.   What error message are you seeing?

coddingtonbear:
Fixed via release 2.0.0.

PR #89, opened 2023-12-20
feat: binary file put support

rowthan:
put file by api not only plain text


docs/openapi.yaml
```diff
@@ -350,6 +350,13 @@ paths:
                 # This is my document
 
                 something else here
+          'multipart/form-data':
+            schema:
+              type: object
+              example: |
+                {
+                  "file": [binary file object]
+                }
```
coddingtonbear: If I understand what you're accomplishing here with this, it's that you're using a combination of the `multipart/form-data` content type when used  with a specific field name (`file`) in its payload as a signal that the incoming content is binary data.  I'm guessing you came to this idea because of how old-style HTML forms work for uploading files, but in the modern world there are saner ways of handling this kind of thing -- especially since this itself is a REST API instead of just a bespoke form handling endpoint.

coddingtonbear (commented):
Thanks for posting a pull request!  I'm at the moment pretty sure we already handle incoming binary content _as long as the content-type header_ is set to one we can recognize, so before continuing on with this, I wonder if we could discuss what problem you're experiencing and your use case via #88?

docs/openapi.yaml
rowthan: I tried set the header, but failed.
I know that you have fixed a error when get binary file, but not **put**.
the code [here](https://github.com/coddingtonbear/obsidian-local-rest-api/blob/main/src/requestHandler.ts#L269)  

only the string type  are available.

i don’t think the multipart/form-data is the key.
The point is that this project did not call the official api `writeBinary `.

rowthan:
@coddingtonbear  is there any demo for binary file put ?

coddingtonbear:
You're totally right; it didn't work!  I still don't think that using `multipart/form-data` is the right choice, though, and have here made a few changes to change things such that they just directly accept the payload data according to the specified `Content-Type` header.  Have a look and let me know if you can see a problem here.

src/constants.ts
```diff
@@ -27,8 +27,8 @@ export const ERROR_CODE_MESSAGES: Record<ErrorCode, string> = {
     "Periodic note does not exist for the specified period.",
   [ErrorCode.RequestMethodValidOnlyForFiles]:
     "Request method is valid only for file paths, not directories.",
-  [ErrorCode.TextOrByteContentEncodingRequired]:
-    "Incoming content must be sent with a bytes or text content encoding.  Be sure to set a Content-type header matching application/* or text/*.",
+  [ErrorCode.TextContentEncodingRequired]:
+    "Incoming content must be text data and have an appropriate text/* Content-type header set (e.g. text/markdown).",
```
coddingtonbear, line 31: I've made this change because, after these changes, we'll only be sending this response back when used with the `PATCH` and `PUT` methods, and in both of those cases, we expect that the incoming data be a string.

src/requestHandler.ts
```diff
@@ -1022,9 +1022,7 @@ export default class RequestHandler {
         limit: MaximumRequestSize,
       })
     );
-    this.api.use(
-      bodyParser.raw({ type: "application/*", limit: MaximumRequestSize })
-    );
+    this.api.use(bodyParser.raw({ type: "*/*", limit: MaximumRequestSize }));
```
coddingtonbear, line 1025: I'm not sure why I'd originally restricted using the `raw` body parser to only `application/*` content-types, but here I've loosened this up such that it'll always use the `raw` parser if the incoming content type wasn't processed by an earlier parser.

src/utils.ts
```diff
@@ -60,3 +60,21 @@ export function getSplicePosition(
   }
   return splicePosition;
 }
+
+export function toArrayBuffer(
+  arr: Uint8Array | ArrayBuffer | DataView | object
+): ArrayBufferLike {
+  if (arr instanceof Uint8Array) {
+    return arr.buffer.slice(arr.byteOffset, arr.byteOffset + arr.byteLength);
```
coddingtonbear, line 68: I found this bug while updating the above tests.  Your original implementation was passing the buffer directly to the caller, and I _think_ that was a bug because that buffer will almost certainly be bigger than your incoming content.  Here, we're instead returning a new slice that contains only the bytes our incoming `Uint8Array` that we need.

src/utils.ts
```diff
@@ -60,3 +60,21 @@ export function getSplicePosition(
...
+  if (arr instanceof DataView) {
+    return arr.buffer.slice(arr.byteOffset, arr.byteOffset + arr.byteLength);
+  }
+  if (arr instanceof ArrayBuffer) {
+    return arr;
+  }
+  // If we've made it this far, we probably have a
+  // parsed JSON object
+  const encoder = new TextEncoder();
+  return encoder.encode(JSON.stringify(arr)).buffer;
```
coddingtonbear, line 79: I'm not super excited about this bit here, but we need some kind of handling for when the incoming data was parsed as JSON (via the `application/json` content-type), and this seemed to be the most reasonable place.

rowthan:
> You're totally right; it didn't work! I still don't think that using `multipart/form-data` is the right choice, though, and have here made a few changes to change things such that they just directly accept the payload data according to the specified `Content-Type` header. Have a look and let me know if you can see a problem here.

agree totally. the reason why i choose ‘multipart/form-data’ is that I just had not find a better way. if you can made it pass the same arguments for both string and binary. is’s really great

PR #139, opened 2024-11-30
fix: Handle PatchFailed error in PATCH request

7shi:
Currently, PatchFailed errors in PATCH requests are not properly handled, causing uncaught promise rejections:

```
plugin:obsidian-local-rest-api:57202 Uncaught (in promise) PatchFailed
    at applyPatch2 (plugin:obsidian-local-rest-api:57202:19)
    at RequestHandler.eval (plugin:obsidian-local-rest-api:57821:60)
    at Generator.next (<anonymous>)
    at fulfilled (plugin:obsidian-local-rest-api:70:24)
```

This PR adds error handling to properly catch and return these errors to the client, including the specific reason for patch failures in the API response.

The change wraps the `applyPatch` call in a try-catch block and uses the existing `returnCannedResponse` method to return a consistent error format.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -38,6 +38,7 @@ export const ERROR_CODE_MESSAGES: Record<ErrorCode, string> = {
   [ErrorCode.MissingOperation]: "No 'Operation' header was provided.",
   [ErrorCode.InvalidOperation]:
     "The 'Operation' header you provided was invalid.",
+  [ErrorCode.PatchFailed]: "Application of this patch has failed.",
 };
 
 export enum ContentTypes {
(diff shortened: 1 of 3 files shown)


Shtooth:
Hello guys!
Any update on this? 
It is really problematic to develop something that uses this as you need to completely restart the API each time you fail a request.

coddingtonbear:
Sorry-- I have very limited time right now owing to it being the holidays, but I imagine I'll have a chance to look at this next week.

coddingtonbear (approved):
Thanks for the patch, @7shi, and sorry for the delay!

PR #151, opened 2025-04-30
modification so that a GET request to “/openapi.yaml” returns the openapi.yaml

spammenotinoz:
Increased compatibility with OpenAPI applications.
Allows direct integration with projects such as https://ObsidianAITools.com via the Obsidian-Local-Rest-API.
These projects expect either an openapi.yaml or openapi.json to be served.

### DIFF
--- a/src/requestHandler.ts
+++ b/src/requestHandler.ts
@@ -21,6 +21,7 @@ import responseTime from "response-time";
 import queryString from "query-string";
 import WildcardRegexp from "glob-to-regexp";
 import path from "path";
+import axios from "axios"; // Add this import at the top
 import {
   applyPatch,
   ContentType,
@@ -255,10 +256,10 @@ export default class RequestHandler {
       certificateInfo:
         this.requestIsAuthenticated(req) && certificate
           ? {
-              validityDays: getCertificateValidityDays(certificate),
-              regenerateRecommended:
-                !getCertificateIsUptoStandards(certificate),
-            }
+            validityDays: getCertificateValidityDays(certificate),
+            regenerateRecommended:
+              !getCertificateIsUptoStandards(certificate),
+          }
           : undefined,
       apiExtensions: this.requestIsAuthenticated(req)
         ? this.apiExtensions.map(({ manifest }) => manifest)
@@ -1258,6 +1259,20 @@ export default class RequestHandler {
 
     this.api.get(`/${CERT_NAME}`, this.certificateGet.bind(this));
     this.api.get("/", this.root.bind(this));
+
+    // Serve /openapi.yaml from remote URL for compatibility
+    this.api.get("/openapi.yaml", async (req, res) => {
+      try {
+        const response = await axios.get(
+          "https://coddingtonbear.github.io/obsidian-local-rest-api/openapi.yaml",
+          { responseType: "text" }
+        );
+        res.setHeader("Content-Type", "application/yaml");
+        res.send(response.data);
+      } catch (err) {
+        res.status(404).send("openapi.yaml not found");
+      }
+    });
 
     this.api.use(this.apiExtensionRouter);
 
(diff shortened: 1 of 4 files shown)


src/requestHandler.ts
```diff
@@ -1256,6 +1257,19 @@ export default class RequestHandler {
...
+      const filePath = path.join(__dirname, "../docs/openapi.yaml");
+      fs.readFile(filePath, "utf8", (err, data) => {
+        if (err) {
+          res.status(500).send("Error reading openapi.yaml");
+        } else {
+          res.set("Content-Type", "application/yaml");
+          res.send(data);
+        }
+      });
+    });
```
coddingtonbear: Unfortunately, I _believe_ this will only work in development (i.e. when you've cloned the repository directly for use in Obsidian); once the package is bundled for a release, the `main.js`, `styles.css` and `manifest.json` are all that actually exist.

That'll mean this has to be approached in a different way, but I have some ideas.  You could, for example, add yaml support to the bundler, then import the `openapi.yaml` file as a module, then return that imported content directly, maybe.

spammenotinoz: > Unfortunately, I _believe_ this will only work in development (i.e. when you've cloned the repository directly for use in Obsidian); once the package is bundled for a release, the `main.js`, `styles.css` and `manifest.json` are all that actually exist.
> 
> That'll mean this has to be approached in a different way, but I have some ideas. You could, for example, add yaml support to the bundler, then import the `openapi.yaml` file as a module, then return that imported content directly, maybe.

Hello, thank-you for the consideration and feedback, I have proposed an alternate approach serving remotely from coddingtonbear.

src/requestHandler.ts
```diff
@@ -1259,6 +1260,20 @@ export default class RequestHandler {
...
+        const response = await axios.get(
+          "https://coddingtonbear.github.io/obsidian-local-rest-api/openapi.yaml",
+          { responseType: "text" }
+        );
+        res.setHeader("Content-Type", "application/yaml");
+        res.send(response.data);
+      } catch (err) {
+        res.status(404).send("openapi.yaml not found");
+      }
+    });
```
coddingtonbear: Unfortunately, this introduces a different kind of problem. Imagine I release a new version of Obsidian Local REST API tomorrow (let’s say v4.0), but you—like many plugin users—haven’t updated and are still running v3.1.0. With the current approach, when an API consumer asks your (v3.1.0) version of Obsidian Local REST API what endpoints it supports, it will mistakenly respond with the capabilities of v4.0—even though those aren’t actually available in your version.

This isn’t a far-fetched scenario. In practice, many users don’t keep plugins fully up-to-date, and mismatches like this are likely to happen. In that context, serving an openapi.yaml file that doesn’t accurately reflect the installed version could lead to more confusion and breakage than not serving one at all.

To avoid this kind of version skew, I think the most reliable approach is to include the correct openapi.yaml file as part of the plugin bundle itself. That way, what gets served will always correspond to the actual running version of the plugin, eliminating any ambiguity. While this might require a small adjustment to the build setup, it's probably the only viable way to ensure long-term correctness.

spammenotinoz: Perhaps I should leave to the experts, this revised approach imports the openapi.yaml during the build.

coddingtonbear: Awesome, thanks, that's getting closer -- I think all that's necessary to make that work are the necessary changes to the bundler configuration.  You can see our bundler configuration in `esbuild.config.mjs`; I'm not exactly sure what changes you'll need to make, but I'm confident you'll figure it out!

spammenotinoz:
Hi I added the below to the build config and the optional definition.
	loader: {
      ".yaml": "text", // Added to handle handle .yaml files as text
    },

coddingtonbear:
Awesome, I think this'll work great!  Thanks for working through the problem carefully and finding a nice, workable solution!

PR #168, opened 2025-06-16
Added the capability to Replace the Header with PATCH. 

taylor-hutyra:
not alot going on here. Just an optional parameter through the API call into _vaultPatchV3 with a check that it is equal to true.
`const replaceHeading = req.get("Replace-Heading") == "true";`

Undefined defaults to false. 

This get's passed on through the PatchInstruction object.

New Example Curl:

curl -X 'PATCH' \
  'http://127.0.0.1:27123/vault/Simple%20Note.md' \
  -H 'accept: */*' \
  -H 'Operation: replace' \
  -H 'Target-Type: heading' \
  -H 'Target-Delimiter: ::' \
  -H 'Target: This is a super simple note::Updating the Header' \
  -H 'Trim-Target-Whitespace: true' \
  -H 'Authorization: Bearer cc64dd3f42437913877c8c942e53b34fb7868def6318aab360d0f586a1087ae5' \
  -H 'Content-Type: text/markdown' \
  -H 'Replace-Heading: true' \
  -d '## New Test Header
Updating header and text
'
Notice the newline between the Header and the Content.

OF NOTE: The package.json currently sets the dependency of markdown-patch to my forked version so that you can test it without accepting the markdown-patch changes first. 
If all is well, it needs to updated to point back to the npm version.


coddingtonbear:
Just to note for those following along, this is dependent upon https://github.com/coddingtonbear/markdown-patch/pull/2, but that's still in the review process.

coddingtonbear:
Closing due to this PR appearing to be abandoned.  Please reach out if you're open to continuing work on this, and I'd be glad to re-open.

PR #173, opened 2025-07-02
feat: Add secure file rename and move operations to PATCH endpoint

duquesnay:
## Summary
Adds secure file rename and move operations to the existing PATCH endpoint, enabling users to reorganize their vault files via the REST API while maintaining link integrity and history.

## Motivation
Users need the ability to programmatically reorganize their vault files while preserving internal links and file history. This extends the existing PATCH endpoint with semantic file operations that integrate seamlessly with Obsidian's file management system.

## Features Added
- **File Rename**: Rename files within the same directory using `Operation: rename` and `Target: name`
- **File Move**: Move files between directories using `Operation: move` and `Target: path`  
- **Link Preservation**: Automatically updates all internal links when files are moved/renamed
- **Directory Creation**: Automatically creates parent directories for move operations
- **Comprehensive Security**: Enterprise-grade validations against various attack vectors

## Security Features
- ✅ Path traversal protection (blocks `../` and absolute paths)
- ✅ Filename validation (blocks invalid characters and reserved names)
- ✅ Length limits and format restrictions  
- ✅ Conflict detection (409 responses for existing destinations)
- ✅ Cross-platform compatibility checks

### Rename a file:
```http
PATCH /vault/folder/old-name.md
Operation: rename
Target-Type: file
Target: name
Content-Type: text/plain

new-name.md
```

### Move a file:
```http
PATCH /vault/folder1/file.md
Operation: move
Target-Type: file  
Target: path
Content-Type: text/plain

folder2/subfolder/file.md
```

## Testing
- 60 tests total (17 new security-focused tests)
- Comprehensive coverage of all validation scenarios
- Edge case testing for malicious inputs
- All existing functionality remains unaffected

## Documentation
- Updated OpenAPI documentation with security restrictions
- Clear examples for both rename and move operations
- HTTP status codes documented for all error scenarios

## Implementation Details
- Uses Obsidian's FileManager API to preserve history and update links
- Validates all inputs before processing to prevent security issues
- Maintains full backward compatibility with existing PATCH operations
- Follows existing code patterns and conventions

## Backward Compatibility
All changes maintain full backward compatibility with existing PATCH operations. No breaking changes to the API.

Ready for review\! Happy to address any feedback or make adjustments as needed.


.github/workflows/test.yml
```diff
@@ -7,9 +7,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-node@v2
       with:
-        node-version: '14'
-    - name: NPM version
-      run: npm --version
+        node-version: '18'
```
coddingtonbear, line 10: It's true that this value should be updated, but can you keep your PR narrowly-scoped to just the very thing you're trying to update?

.gitignore
```diff
@@ -17,3 +17,29 @@ main.js
...
+*.log
+
+# Development documentation
+AGENT_TASK.md
+HOT-RELOAD-ANALYSIS.md
+RELOAD-INSTRUCTIONS.md
+SCRATCHPAD.md
+SPIKE-*.md
+fix-hot-reload.md
+hot-reload-fix.md
```
coddingtonbear, line 45: I think this is something leftover from your local environment.

.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 18.18.0
```
coddingtonbear, line 1: Again, please try to keep your changes narrowly-scoped to the feature you're proposing.

esbuild.config.mjs
```diff
@@ -51,5 +51,8 @@ esbuild
     sourcemap: prod ? false : "inline",
     treeShaking: true,
     outfile: "main.js",
+	loader: {
+      ".yaml": "text", // Added to handle handle .yaml files as text
+    },
```
coddingtonbear, line 56: I'm vaguely aware that this is _already_ in `main`; so I'm not clear on why this is showing a diff; any ideas?

coddingtonbear (changes requested):
Hi there—thanks for taking the time to put this together. I’m excited about the idea of supporting rename and move operations (handling them via PATCH makes a lot of sense!).

I’m running into trouble reviewing the PR in its current form, though: the diff shows about 15 000 lines, including many changes that already exist on main. To help me give you a meaningful review, could you please:

1. Confirm your branch/PR settings (it looks as if some commits from main may have slipped in).
2. Scope the PR to a single concern. Ideally that would mean one PR for “rename” and another for “move,” each limited to the code and tests directly related to that feature.

Project-wide tweaks—changing Node versions, updating .tool-versions, modifying .gitignore, etc.—are valuable, but they’re best handled in their own small PRs so they don’t hide the feature changes.

A quick heads-up: every new feature carries an ongoing maintenance cost, so I can’t always accept big, complex additions. In future, feel free to start a discussion before diving in so we can confirm the approach first. And remember, if something doesn’t fit the core project, you can always add new routes through a plugin extension (see the wiki: https://github.com/coddingtonbear/obsidian-local-rest-api/wiki/Adding-your-own-API-Routes-via-an-Extension).

Thanks again for contributing—once the PR is trimmed down, I’ll be happy to take another look!

esbuild.config.mjs
duquesnay: No idea. It is indeed in the main, added 2 months ago. I tried to pull from the latest main version, but might have a mix up here. Anyway, abandoning this PR for a cleaner one

PR #176, opened 2025-07-08
feat: Add file rename functionality via PATCH /vault/{filepath}

duquesnay:
## Summary
adds file rename functionality to the Obsidian Local REST API

## API Usage

Rename a file using PATCH with these headers:
```bash
curl -X PATCH https://localhost:27124/vault/path/to/file.md \
  -H "Operation: replace" \
  -H "Target-Type: file" \
  -H "Target: name" \
  -d "new-filename.md"
```

## Implementation Details

- **API Endpoint**: PATCH `/vault/{filepath}` 
- **Headers Required**: 
  - `Operation: replace` (legacy compatibility)
  - `Target-Type: file` 
  - `Target: name`
- **Request Body**: New filename
- **Link Preservation**: Uses `app.fileManager.renameFile()` to update internal vault references
- **Error Handling**: Returns 404 for missing files, 409 for conflicts, 400 for validation errors

## Commit Structure

This PR contains 5 commits extracted from git history:

1. **3221a95** - `fix: Add null check for Target header in PATCH requests`
2. **7bd91d8** - `fix: improve error messages for PATCH operations` 
3. **77eac7a** - `feat: Add file rename endpoint with PATCH /vault/{filepath}`
4. **4636d36** - `test: Add tests for file rename operation`
5. **b7702b3** - `docs: Add OpenAPI documentation for file rename operation`

## Testing

Includes tests for:
- Successful rename operation  
- Error handling for non-existent files
- Conflict detection when destination exists
- FileManager API integration via mocks

## Files Modified

- `src/requestHandler.ts` - Core implementation and fixes
- `src/requestHandler.test.ts` - Test coverage
- `src/constants.ts` - Error message improvements
- `docs/openapi.yaml` - API documentation
- `docs/src/lib/patch.jsonnet` - Documentation templates

### DIFF
--- a/src/requestHandler.ts
+++ b/src/requestHandler.ts
@@ -509,6 +509,14 @@ export default class RequestHandler {
       });
       return;
     }
+    
+    // Check for file-level operations (like rename) BEFORE validation
+    if (targetType === "file" && operation === "replace") {
+      if (rawTarget === "name") {
+        return this.handleRenameOperation(path, req, res);
+      }
+    }
+    
     if (!["heading", "block", "frontmatter"].includes(targetType)) {
       this.returnCannedResponse(res, {
         errorCode: ErrorCode.InvalidTargetTypeHeader,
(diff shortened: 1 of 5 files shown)


src/requestHandler.ts
```diff
@@ -509,6 +509,14 @@ export default class RequestHandler {
       });
       return;
     }
+    
+    // Check for file-level operations (like rename) BEFORE validation
+    if (targetType === "file" && operation === "replace") {
+      if (rawTarget === "name") {
+        return this.handleRenameOperation(path, req, res);
+      }
+    }
```
coddingtonbear, line 518: If you look at theother Operations and Target types we support:

- Operations: `append`, `prepend`, `replace`
- Target-type: `heading`, `block`, `frontmatter`

All of those are universally valid in any combination (i.e. you can use `append` with `block` and any other Target type, and that remains true for any Operation or Target-type you pick).

This here is very different from that -- you can use `rename` only when you've picked the operation of `filename` and a `target` of "name", and you can use `filename` only when you've picked `rename` and a `target` of `name` and so on.  This new operation, target, and target type are valid *only* in combination and only when used exactly as arrayed above.

This is making me think that we should take some effort to try to find away of making this new API a little more cohesive with the original one, *or* create an entirely separate API for supporting this kind of thing.

I'm not exactly sure what shape this should take, so I had a brief conversation with an LLM to try to see if it could come up with any clever solutions, and it came up with *some* ideas (see https://chatgpt.com/share/688fa27b-7b90-800f-a1b8-085cea5c07ad ), but I don't know that I'm all that excited about those, either.

I'm interested to hear what you think!

coddingtonbear (changes requested):

src/requestHandler.ts
duquesnay: I'm up to discuss this. 
To be honest, I have probably the equivalent of 20PR on my own fork: https://github.com/duquesnay/obsidian-local-rest-api. I needed more features to be able to connect an MCP (AI connector) so I have now a pretty extensive list of features. 
I'm recreating these PR here by extracting things in a more atomic way (not the best process, I'm new at contributing).
Maybe have a look and tell me what you think about the API choices?

coddingtonbear: As much as I wish I could, I'm afraid I really can't spare the time to review a whole project. It's really very time-consuming to review LLM-generated code, and there's no way I'll be able to fit that in the time I've budgeted for spending on this project monthly.

What I'm going to recommend to you, going forward, is that you first start by making your API extensions via the extension process documented here: https://github.com/coddingtonbear/obsidian-local-rest-api/wiki/Adding-your-own-API-Routes-via-an-Extension.

For the moment, I think I'm going to close this PR, but I look forward to seeing your API extension if you end up making that effort!  Cheers & good luck!

PR #177, opened 2025-07-08
Add File Move Operation to PATCH Endpoint

duquesnay:
## Summary
Adds file move operation to the PATCH endpoint, allowing files to be moved within the vault while preserving internal links.

## Dependencies
- Depends on PR #176 (file rename) - must be merged first

## Changes
- Add `move` operation to PATCH endpoint for files
- Implement `handleMoveOperation` method
- Add path traversal security validation
- Add 8 test cases covering success and error scenarios
- Update OpenAPI documentation
- Refactor PATCH validation logic

## Security
- Blocks path traversal attempts (`../`)
- Rejects absolute paths
- Validates and normalizes all input paths

## Usage
```bash
curl -X PATCH https://localhost:27124/vault/old/path/file.md \
  -H "Operation: move" \
  -H "Target-Type: file" \
  -H "Target: path" \
  -H "Content-Type: text/plain" \
  -d "new/location/file.md"
```

## Implementation
- Uses `FileManager.renameFile()` to preserve links
- Creates parent directories as needed
- Returns appropriate error codes and messages
- Follows existing error handling patterns

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -279,7 +279,7 @@ paths:
         your data as JSON (particularly when appending, for example,
         list items).
         
-        ## File Rename Operation
+        ## File Operations
         
         ### Renaming a File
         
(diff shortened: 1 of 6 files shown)


coddingtonbear (changes requested):
I think this PR might be a good solution for the concerns I had around #176 so I think this is worth talking over more!  

This PR, though, looks like it includes all of the changes of *both* #176 and this PR rather than being just the unique work for this particular problem.  To make this PR reviewable, what you need to do is set its base branch to be the branch you made for #176 rather than `main`.

Issue #172, opened 2025-06-23
Inconsistent search results when using /search/simple


kurokirasama:
Hey, for some reason not all notes are found succefully, despite earhing for the exact name and subdirectory.

coddingtonbear:
I'm afraid I can't do anything without specific details about your requests including what documents you were expecting to find but were not; can you provide that insight?

kurokirasama:
no problem. For instance, i have several notes with with word "prompt" in their title. But when searching only a few appear (actually just 4)...

I mean, in the actual search results (search/simple) there are a lot of notes, but if they i filter by filename, with the word promp in it, only 4 remains.

I hope to be clear.

coddingtonbear:
I'm afraid I'll need more than that -- can you attach the actual files you're not finding as well as the actual request you're making? Without that, there's no test case I can set up.

You can use `httpie`'s `-v` option to have it print the raw request and response.

kurokirasama:
is curl -v -i also useful? or do you need httpie?

coddingtonbear:
Just to make sure it's clear, I'll still need the exact file (and its filename) of the one you're not consistently finding in search results.

kurokirasama:
i have a note titled (and therefore the filename): prompt fix bug (prompt fix bug.md), with content:

```md
## INFORMATION
## CODE
## ERROR
```

that does not appear in the search results

coddingtonbear:
Great, I think that's enough information for somebody to at least try to see if your issue can be reproduced.

It might be a few weeks before I can find the time to do this investigation myself, but if anybody else comes across this issue before I can get to it and is able to reproduce this problem, or, better yet, post a PR fixing it, that would be amazing.

coddingtonbear:
I'm afraid I haven't been able to reproduce any inconsistencies in my tests around this.  Here's my working example:

1. I've uploaded this file and stored it in my vault as `issue_172.md`:  [issue_172.md](https://github.com/user-attachments/files/21337622/issue_172.md)
2. I run the following `httpie` command repeatedly: `http -v --verify=no --auth-type=bearer --auth="REDACTED" POST http://127.0.0.1:27123/search/simple/\?query\=f486e851-f665-413b-aa9f-faacff6b7a36`
3. I see exactly the following result for each response:
 
```[
    {
        "filename": "issue_172.md",
        "matches": [
            {
                "context": "",
                "match": {
                    "end": 100,
                    "start": 64
                }
            }
[...]

PR #187, opened 2025-09-17
Include the title of the file in the simple search

phortx:
... also some smaller formatting fixes that the IDE did.

See #172

### DIFF
--- a/mocks/obsidian.ts
+++ b/mocks/obsidian.ts
@@ -22,7 +22,7 @@ class DataAdapter {
   _read = "";
   _readBinary = new ArrayBuffer(0);
   _write: [string, string];
-  _writeBinary : [string, ArrayBuffer];
+  _writeBinary: [string, ArrayBuffer];
   _remove: [string];
   _stat = new Stat();
 
@@ -42,12 +42,12 @@ class DataAdapter {
     return this._readBinary;
   }
 
-  async write(path: string, content: string, option?:DataWriteOptions): Promise<void> {
+  async write(path: string, content: string, option?: DataWriteOptions): Promise<void> {
     this._write = [path, content];
   }
 
-  async writeBinary(path: string, content: ArrayBuffer, option?:DataWriteOptions): Promise<void> {
-    this._writeBinary = [path,content]
+  async writeBinary(path: string, content: ArrayBuffer, option?: DataWriteOptions): Promise<void> {
+    this._writeBinary = [path, content]
   }
 
   async remove(path: string): Promise<void> {
@@ -72,7 +72,7 @@ export class Vault {
     return this._cachedRead;
   }
 
-  async createFolder(path: string): Promise<void> {}
+  async createFolder(path: string): Promise<void> { }
 
   getFiles(): TFile[] {
     return this._files;
@@ -158,6 +158,7 @@ export class FileStats {
 
 export class TFile {
   path = "somefile.md";
+  basename: string = "somefile";
   stat: FileStats = new FileStats();
 }
 
@@ -165,7 +166,7 @@ export class PluginManifest {
   version = "";
 }
 
-export class SettingTab {}
+export class SettingTab { }
 
 export const apiVersion = "1.0.0";
 
--- a/src/requestHandler.ts
+++ b/src/requestHandler.ts
@@ -262,10 +262,10 @@ export default class RequestHandler {
       certificateInfo:
         this.requestIsAuthenticated(req) && certificate
           ? {
-              validityDays: getCertificateValidityDays(certificate),
-              regenerateRecommended:
-                !getCertificateIsUptoStandards(certificate),
-            }
+            validityDays: getCertificateValidityDays(certificate),
+            regenerateRecommended:
+              !getCertificateIsUptoStandards(certificate),
+          }
           : undefined,
       apiExtensions: this.requestIsAuthenticated(req)
         ? this.apiExtensions.map(({ manifest }) => manifest)
@@ -997,7 +997,7 @@ export default class RequestHandler {
 
     for (const file of this.app.vault.getMarkdownFiles()) {
       const cachedContents = await this.app.vault.cachedRead(file);
-      const result = search(cachedContents);
+      const result = search(file.basename + "\n\n" + cachedContents);
       if (result) {
         const contextMatches: SearchContext[] = [];
         for (const match of result.matches) {


src/requestHandler.ts
```diff
@@ -997,7 +997,7 @@ export default class RequestHandler {
 
     for (const file of this.app.vault.getMarkdownFiles()) {
       const cachedContents = await this.app.vault.cachedRead(file);
-      const result = search(cachedContents);
+      const result = search(file.basename + "\n\n" + cachedContents);
```
coddingtonbear: ~~I believe the filename is already included in the response; see line 1017 below.  Am I misunderstanding the result shape, or was there a reason that wasn't sufficient?~~

Nevermind, I see that this is instead adding this to the searched text.  I'll think about this a little more when I have more time.

phortx: Yes, see my last comments in the issue above :) 

The problem is that searching for the file name doesn't yield matching files with that name. However this is important.

My change makes sure the filename is also respected in the search.

Thanks, take your time and let me know when you need anything :)

coddingtonbear: I think it's a great idea to make `/search/simple/` also match filenames, but there are a couple things we'll need to work together to fix before this could land in `main`:

1. If you look below around line 1006,1007, you'll see that we include the `match` position within the document in our response.  Before these changes, the document we're searching within *is* the content of the markdown file, and folks could be using those returned `start`/`end` positions to figure out where to make changes to the document. After your changes, though, that returned position no longer matches the position of the match within the document, and if they have some kind of script for making changes based on where the match is found, suddenly their tooling will be modifying the wrong position in the document.
2. Related to that, there isn't really a match range to give if you've matched the file's basename.  Maybe instead of just returning `match: {start, end}`, you could change the API such that it returned something like `match: {start, end, source}` where `source` could be either `content` or `basename`?

I'm curious to hear what you think about either of those

phortx: Thanks for pointing this out! 

1. Should be easy to fix by just adding 3 to the start and end position (first line is headline, then 2 linebreaks)
2. That's a good idea, or we return 0 for start and end and document that this means, the match was in the headline.

I will look into this as soon as I have some spare time :)

phortx: Sorry for the delay, busy times :)

I made the changes, could you please take a look when you have some time? Thanks!

phortx: I also discovered a bug with the `contextLength`, which caused the search result often to contain no context. That bug also exists in your main branch.

Details: When `parseInt()` fails, it returns `NaN` (instead of `null`). But the Nullish Coalescing Operator (??) only respects `null` or `undefined`. Thus `contextLength` was `null` most of the times and therefore the `context` field empty. I fixed that.

phortx:
Hi @coddingtonbear,

what do you think? Could we merge this by time? Do you need anything from my side? :)

src/requestHandler.ts
```diff
@@ -1012,20 +1012,44 @@ export default class RequestHandler {
...
+                start: match[0] - positionOffset,
+                end: match[1] - positionOffset,
+                source: "content"
+              },
+              context: cachedContents.slice(
+                Math.max(match[0] - contextLength, positionOffset),
+                match[1] + contextLength
+              ),
+            });
+          }
```
coddingtonbear: There's a lot of possibility for off-by-one types of errors and other kinds of bugs in this kind of logic; could I trouble you to writing some tests covering this?

phortx: Thanks, I really appreciate your in-depth review. I will add more tests in the next days. Have a great sunday!

phortx: Update: I fixed all copilot comments in the MR. It found a bug, which is nice!

Also I added lot's of test cases. Tested: works like a charm:

```bash
$ curl -X 'POST' \
      'http://127.0.0.1:27123/search/simple/?query=Master%20Plan' \
      -H 'accept: application/json' \
      -H 'Authorization: Bearer 320fac56a929e907fd1f724b12eb55327b38a363e8c8a8cbc290c9c5f1c30631'
      
[
  [...],
  {
    "filename": "5 - Finanzen/1 - Master Plan.md",
    "score": -1.2066,
    "matches": [
      {
        "match": {
          "start": 4,
          "end": 10,
          "source": "filename"
        },
        "context": "1 - Master Plan"
      },
      {
        "match": {
          "start": 11,
          "end": 15,
          "source": "filename"
        },
        "context": "1 - Master Plan"
      }
    ]
  }
]
```

mocks/obsidian.ts
```diff
@@ -177,5 +178,24 @@ export class SearchResult {
 export function prepareSimpleSearch(
   query: string
 ): (value: string) => null | SearchResult {
-  return null;
+  return (text: string) => {
```
coddingtonbear: Just a quick note about this file, @phortx: since it lives under `mocks/`, it’s meant purely as a mock—basically something lightweight so the type checker and tests have a stand-in. We shouldn’t put any real implementation logic here.

Instead, the idea is that your tests can _control_ what this mock returns in different scenarios. If you take a peek at some of the other mock functions in `mocks/obsidian.ts`, you’ll see how they’re set up to let tests override behavior as needed.

So to make this consistent: this function shouldn’t implement any real behavior. It should either do nothing or return whatever value your test explicitly configures, so that you can verify how `requestHandler.ts` responds to those different return values.

phortx: Good point, I resetted the mock implementation and added a way for the tests to override the behavior.

src/requestHandler.ts
```diff
@@ -1012,20 +1012,44 @@ export default class RequestHandler {
...
+          if (match[0] < positionOffset) {
+            // When start position is between 0 and positionOffset, that means the search term matched within the filename.
+            contextMatches.push({
+              match: {
+                start: match[0],
+                end: match[1],
+                source: "filename"
+              },
+              context: file.basename,
+            });
```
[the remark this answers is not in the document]
phortx: fixed

src/requestHandler.ts
```diff
@@ -1012,20 +1012,44 @@ export default class RequestHandler {
...
       const cachedContents = await this.app.vault.cachedRead(file);
-      const result = search(cachedContents);
+
+      // Add the headline to the search text to include it in the search.
+      const filenamePrefix = file.basename + "\n\n";
+      const result = search(filenamePrefix + cachedContents);
+
+      // We added the headline to the search text with 2 line breaks.
+      // That causes the start and end position numbers to be wrong with an offset
+      // of the char length of the headline line breaks.
```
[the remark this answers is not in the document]
phortx: fixed

mocks/obsidian.ts
```diff
@@ -42,12 +42,12 @@ class DataAdapter {
...
 
-  async write(path: string, content: string, option?:DataWriteOptions): Promise<void> {
+  async write(path: string, content: string, option?: DataWriteOptions): Promise<void> {
     this._write = [path, content];
   }
 
-  async writeBinary(path: string, content: ArrayBuffer, option?:DataWriteOptions): Promise<void> {
-    this._writeBinary = [path,content]
+  async writeBinary(path: string, content: ArrayBuffer, option?: DataWriteOptions): Promise<void> {
+    this._writeBinary = [path, content]
```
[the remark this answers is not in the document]
phortx: fixed

src/requestHandler.ts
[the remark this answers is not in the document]
phortx: fixed

src/requestHandler.ts
```diff
@@ -997,8 +997,8 @@ export default class RequestHandler {
         errorCode: ErrorCode.InvalidSearch,
       });
     }
-    const contextLength: number =
-      parseInt(req.query.contextLength as string, 10) ?? 100;
+    const contextLengthRaw = parseInt(req.query.contextLength as string, 10);
+    const contextLength = Number.isNaN(contextLengthRaw) ? 100 : contextLengthRaw;
```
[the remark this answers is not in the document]
coddingtonbear: @Co-pilot is being a little too paranoid here; let's ignore this one.

docs/openapi.yaml
```diff
@@ -1583,11 +1583,17 @@ paths:
                             properties:
                               end:
                                 type: "number"
+                                description: "End character position of the match within the file. Starts with 0."
                               start:
                                 type: "number"
+                                description: "Start character position of the match within the file. Starts with 0."
```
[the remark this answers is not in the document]
coddingtonbear: I don't quite understand what @co-pilot is going for here, but I think the original text is just fine!

src/requestHandler.test.ts
```diff
@@ -802,4 +804,291 @@ describe("requestHandler", () => {
...
+        .expect(401);
+    });
+
+    test("missing query parameter", async () => {
+      await request(server)
+        .post("/search/simple/")
+        .set("Authorization", `Bearer ${API_KEY}`)
+        .expect(400);
+    });
+  });
```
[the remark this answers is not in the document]
coddingtonbear: :shrug:

phortx:
> Hey there @phortx, we're almost there! Could you comment on the remaining @co-pilot suggestions above? Once we have @co-pilot happy, I'll do my own review and get this merged.

Yes, will take a look in the next days when it's getting more relaxed here :) Thanks!

src/requestHandler.ts
```diff
@@ -1012,20 +1012,48 @@ export default class RequestHandler {
...
-              end: match[1],
-            },
-            context: cachedContents.slice(
-              Math.max(match[0] - contextLength, 0),
-              match[1] + contextLength
-            ),
-          });
+          // Check if the entire match is within the filename (including the newlines).
+          // We need to ensure both start and end positions are within the filename prefix.
+          if (match[0] < positionOffset && match[1] <= positionOffset) {
```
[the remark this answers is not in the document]
phortx: Interesting edge case. I changed the behavior to ignore matches that span the boundary between filename and content (who searches intentionally `filename\n\ncontent` anyway?). I think that's the simplest and most robust solution. Also added an test for that case.

phortx:
Fixed everything! Could you take a look?

src/requestHandler.ts
```diff
@@ -1012,20 +1012,51 @@ export default class RequestHandler {
...
+            // When start position is between 0 and positionOffset and end position is <= positionOffset,
+            // that means the search term matched entirely within the filename prefix.
+            // Clamp the end position to the basename length to ensure it doesn't exceed the context.
+            contextMatches.push({
+              match: {
+                start: match[0],
+                end: Math.min(match[1], file.basename.length),
+                source: "filename"
```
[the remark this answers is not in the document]
coddingtonbear: Now, now, @copilot; relax.

coddingtonbear:
**Congratulations, @phortx, you made it!** 🎉

Thanks for sticking with this through the back-and-forth. Even for relatively small changes, I have to be pretty careful about what lands here—since I’m the sole maintainer, anything that gets merged becomes something I need to understand and be able to maintain and debug long-term.

I really appreciate you being willing to iterate until the solution was clear and solid. That kind of collaboration makes it much easier for me to merge contributions with confidence. Thanks again for the effort!

phortx:
@coddingtonbear thanks for your kind words and your help on this topic! It was a pleasure to work with you :) 

You did everything right: It's your responsibility to understand the code and ensure it's stable, maintainable and working well. That's the burden of the maintainer. You have to act carefully and take your time. So you acted like a responsible maintainer. Well done! ;)

PR #195, opened 2025-11-08
Add rendered HTML content support via Accept header

papertray3:
## Summary

This PR adds support for retrieving **rendered HTML content** from markdown files using HTTP content negotiation. Clients can now request fully-rendered HTML by specifying `Accept: application/vnd.olrapi.note+html` in GET requests.

## Changes

- **New Content Type**: Added `ContentTypes.olrapiNoteHtml` constant (`application/vnd.olrapi.note+html`)
- **HTML Rendering**: Implemented `renderMarkdownToHtml()` method using Obsidian's native `MarkdownRenderer.render()` API
- **Request Handling**: Updated `_vaultGet()` to handle HTML content type requests
- **Documentation**: Updated OpenAPI specification with new content format examples
- **Memory Management**: Proper component lifecycle management to prevent memory leaks

## Features

The rendered HTML output includes:
- Wiki-links converted to HTML anchors with proper `data-href` attributes
- Obsidian-specific markdown (callouts, embeds, transclusions)
- Plugin integrations preserved (e.g., Metadata Menu icons)
- Syntax-highlighted code blocks
- Full semantic HTML structure

## API Usage

```bash
# Get rendered HTML
curl -k https://127.0.0.1:27124/vault/path/to/note.md \
  -H "Accept: application/vnd.olrapi.note+html"
```

## Backward Compatibility

This feature:
- Uses standard HTTP content negotiation (Accept header)
- Does not modify existing behavior for markdown or JSON responses
- Works with all GET endpoints (`/vault/`, `/active/`, `/periodic/`)
- Is fully optional - defaults to existing markdown behavior

## Testing

Tested with:
- Various markdown files with wiki-links, embeds, and frontmatter
- Complex documents with Dataview queries
- Plugin integrations (Metadata Menu)
- Both `/vault/{filename}` and `/active/` endpoints

## Use Cases

This feature enables:
- Digital garden publishing workflows with preview capabilities
- External rendering of Obsidian notes
- Wiki-link validation before publishing
- Integration with static site generators

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,3 +16,80 @@ This is particularly useful if you need to interact with Obsidian from a browser
 ## Credits
 
 This was inspired by [Vinzent03](https://github.com/Vinzent03)'s [advanced-uri plugin](https://github.com/Vinzent03/obsidian-advanced-uri) with hopes of expanding the automation options beyond the limitations of custom URL schemes.
+
+## Rendered Content Support
+
+This plugin can extract fully-rendered content from your notes, including:
+
+- **Dataview queries** - Tables, lists, and task queries are fully executed
+- **DataviewJS** - Custom JavaScript blocks are rendered
+- **Plugin-generated content** - Metadata Menu, Buttons, and other plugin outputs
+- **Dynamic dashboards** - Perfect for AI assistants to see computed views
+
+### Content Formats
+
+**Plain Text** (`application/vnd.olrapi.note+rendered-text`)
+- Human-readable text output
+- Fast and simple
+- Good for general text analysis
+
+**Structured JSON** (`application/vnd.olrapi.note+rendered-json`)
+- AI-optimized structured data
+- Tables as 2D arrays with headers
+- Document hierarchy preserved with type tags
+- Complete frontmatter included
+- Perfect for programmatic access
+
+### Example Usage
+
+```bash
+# Get rendered text
+curl -H "Authorization: Bearer YOUR_API_KEY" \
+     -H "Accept: application/vnd.olrapi.note+rendered-text" \
+     "https://localhost:27124/vault/path/to/note.md"
+
+# Get structured JSON
+curl -H "Authorization: Bearer YOUR_API_KEY" \
+     -H "Accept: application/vnd.olrapi.note+rendered-json" \
+     "https://localhost:27124/vault/path/to/note.md"
+```
+
+**JSON Output Structure:**
+```json
+{
+  "metadata": {
+    "sourcePath": "path/to/note.md",
+    "renderedAt": "2025-01-10T18:00:00.000Z",
+    "format": "json",
+    "version": "1.0"
+  },
+  "frontmatter": {
+    "tags": ["example"],
+    "category": "Documentation"
+  },
+  "content": [
+    {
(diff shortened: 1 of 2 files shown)


papertray3:
Follow-up: I discovered Dataview/DataviewJS blocks were still empty unless the renderer had a real DOM host. The latest commit      
  attaches a hidden container to document.body, renders there, waits for dynamic blocks to settle, then removes it. This ensures the HTML
  returned by /vault matches what Obsidian shows (full Workbench/Queue/Vault stats, etc.). Added coverage for the Accept: application/   
  vnd.olrapi.note+html path while I was here.

README.md
```diff
@@ -16,3 +16,80 @@ This is particularly useful if you need to interact with Obsidian from a browser
 ## Credits
 
 This was inspired by [Vinzent03](https://github.com/Vinzent03)'s [advanced-uri plugin](https://github.com/Vinzent03/obsidian-advanced-uri) with hopes of expanding the automation options beyond the limitations of custom URL schemes.
+
+## Rendered Content Support
```
coddingtonbear, line 20: We have public docs in the `/docs/` folder.  Instead of documenting this in the readme, that should probably be in the docs themselves if you could.

RENDERED_CONTENT.md
```diff
@@ -0,0 +1,540 @@
+# Rendered Content API
```
coddingtonbear, line 1: I'm betting this is an artifact created when you asked an LLM to help you code this and I bet you didn't intend to include this in the PR.

src/constants.ts
```diff
@@ -51,6 +59,9 @@ export enum ContentTypes {
   json = "application/json",
   markdown = "text/markdown",
   olrapiNoteJson = "application/vnd.olrapi.note+json",
+  olrapiNoteHtml = "application/vnd.olrapi.note+html",
+  olrapiRenderedText = "application/vnd.olrapi.note+rendered-text",
+  olrapiRenderedJson = "application/vnd.olrapi.note+rendered-json",
```
coddingtonbear, line 64: Forgive me, but this PR is already extremely long -- could I trouble you to break this apart into individual PRs for each of these types of content?  Only *one* of these content types appears to be discussed in this PR; so I think sticking to rendered HTML in this one is probably the right choice, and I want to say: I might need some convincing before I'll be up for taking on a patch for the other two rendering options.

Second thing: you don't need to invent a new content type for things like html/text/json -- those already exist:

- `text/html`
- `text/plain`
- `application/json`

src/main.ts
```diff
@@ -665,10 +695,100 @@ class LocalRestApiSettingTab extends PluginSettingTab {
...
+              const num = parseInt(value);
+              if (!isNaN(num) && num > 0) {
+                this.plugin.settings.renderCacheTimeoutMs = num;
+                await this.plugin.saveSettings();
+              }
+            })
+        );
+
+      // Cache statistics
+      if (this.plugin.renderCacheManager) {
```
coddingtonbear, line 760: Was there a particular performance problem you found that inspired you to want to cache renders?  It seems like kind of a lot of complication to add unless there are severe performance problems.

PR #199, opened 2025-11-12
Added a /Tags endpoint to query all tags

glnds:
## Summary

Added a new endpoint to retrieve all tags from the vault, enabling downstream applications to query the complete tag collection.

## Motivation

This new endpoint allows the MCP Obsidian integration to be extended with access to the full tag inventory for intelligent tag suggestions when creating notes.

## Changes

- Implemented GET endpoint for retrieving all vault tags
- Returns a comprehensive tag list for consumption by external tools

## Impact

This endpoint will enable `mcp-obsidian` to be easily extended to query available tags

The feature is particularly valuable for AI-assisted note creation workflows where tag suggestions improve organization and discoverability.


Obsidian_Local_REST_API.postman_collection.json
```diff
@@ -0,0 +1,455 @@
+{
```
coddingtonbear: At the moment, I won't be able to find the time to verify this collection -- could you separate this out into its own PR so this postman collection and your actual change request can be considered separately?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "obsidian-local-rest-api",
-  "version": "3.2.0",
```
coddingtonbear: Could you remove these version number bumps?  I handle those as part of the release process and will increment the version number following [semver guidelines](https://semver.org/) depending on what kinds of changes end up in that release.

Also, by the way, since this is a new *feature* instead of a *bugfix* this would be a minor version bump instead of a patch (i.e. 3.2.0 -> 3.3.0 instead of 3.2.0 -> 3.2.1).

versions.json
```diff
@@ -54,5 +54,6 @@
 	"3.0.4": "0.12.0",
 	"3.0.5": "0.12.0",
 	"3.1.0": "0.12.0",
-	"3.2.0": "0.12.0"
```
coddingtonbear: & of course this should be removed, too, since I'll generate this as part of the release.

CLAUDE.md
```diff
@@ -0,0 +1,132 @@
+# CLAUDE.md
```
coddingtonbear: I think the CLAUDE.md file was added accidentally. It looks like a temporary artifact produced by the Claude AI assistant, not something intended for the repo. Can you remove it from this PR?

src/requestHandler.ts
```diff
@@ -946,6 +946,14 @@ export default class RequestHandler {
     );
   }
 
+  async tagsGet(req: express.Request, res: express.Response): Promise<void> {
+    const allTags = (this.app.metadataCache as any).getTags();
```
coddingtonbear: If you've had to cast this as `any` to make `.getTags` accessible, it makes me think that maybe the obsidian types are out-of-date?

That being said, I don't see them documented [in the official types](https://github.com/obsidianmd/obsidian-api/blob/master/obsidian.d.ts) -- are you sure this exists and is a public interface?

Makefile
```diff
@@ -0,0 +1,60 @@
+.PHONY: help build install dev clean
```
coddingtonbear: I understand that this might be helpful for you, but I'm not sure I want to take on maintenance of this particular part of this PR and instead would rather we keep all scripts in `package.json`.  Could I trouble you into removing this?

I know that it seems harmless to add unrelated things in a single PR, but unfortunately every bit of code that's added is a liability that will require future maintenance, support, and updates as time goes on, and I really need to be careful about what I agree to maintain going forward.

coddingtonbear (changes requested):
Thanks for the patch! I think this is totally something that could be useful to folks; so I'm up for merging this, but there are a few things I'll need from you first:

1. I added a few comments in here about things that either shouldn't be in a PR or should be in a separate one.
2. Leave the version incrementing to me.  These changes will probably end up bundled into a larger release, and the version number you suspect I'll increment to may not match what it actually turns out to be (e.g. if your changes are bundled with a backward-incompatible change, the *major* version number would be incremented, etc.).
3. I'll need to ask you to update the docs to include this new API endpoint.  See the `docs` folder for a rough idea about what you'll need to do for that.

Once you've had a chance to look into all of this, let me know and I'll have another look!  Thanks again for the patch!

PR #201, opened 2025-11-27
Update release.yml for zip

SuperHappycat:


SuperHappycat:
Can you please merge this?

coddingtonbear:
Thanks for the fix — I’ve just merged this 👍

For context: I review and merge PRs as time allows, and it can sometimes take a while even for small changes. I appreciate your patience here, and thanks again for the contribution!

coddingtonbear:
Just for posterity, in case somebody comes across this issue later: although this did some necessary upgrades to the github action, this did not fix the existing problem we currently have in which the version number does not match the release.

PR #207, opened 2026-01-02
Add POST /vault/move endpoint for moving/renaming files and folders

spudro228:
## Summary
- Added new POST `/vault/move` endpoint for moving and renaming files and folders in the vault
- Supports moving files between directories
- Supports renaming files and folders
- Returns the new path after successful move
- Includes OpenAPI documentation, tests, and proper error handling

## Test plan
- [x] Unit tests added for the new endpoint
- [x] Manual testing with Obsidian vault


coddingtonbear:
Thanks so much for taking the time to put this PR together — I really appreciate the effort you’ve put into it.

After reviewing it, I think this ends up overlapping significantly with the problem being addressed in #191. That issue is aiming to solve the same underlying need, but in a way that fits a bit more cleanly with the existing REST-style approach, so I don’t think I can move this PR forward in its current form.

That said, this definitely doesn’t have to be the end of the road:
- If what you’re specifically interested in is this implementation of a “move”-style operation, you could adapt this work into an API extension instead. The project explicitly supports that model, and it lets you publish and iterate independently without needing upstream approval. There’s some documentation on how to do that here:
https://github.com/coddingtonbear/obsidian-local-rest-api/wiki/Adding-your-own-API-Routes-via-an-Extension
- If your main goal is to see a “move”-type operation supported in the core API, contributions to #191 would be very welcome. I like the direction @mairas has taken there, and it’s fairly close — there are just a few remaining pieces to work through before it’s ready.

I know having a PR declined can be frustrating, and I’m sorry to be the bearer of that news. I try to be direct in these situations so folks have clarity as early as possible. For future work, starting with a quick discussion in the Discussions tab can be a good way to align on approach ahead of time and avoid duplicated effort.

Thanks again for the contribution and for engaging with the project.

PR #208, opened 2026-01-03
Fix openapi generator build fail

Marc-R2:
* Added an explicit `items: {}` property to array type definitions in `docs/src/openapi.jsonnet` to ensure proper JSON Schema compliance for array types. `openapi-generator` build fails without.
* When running 'build-docs', some descriptions changed based on 6d7cb6a72d6d5fb33f4f89f09366cb4ecac18043, which has not yet been reflected in the build. The fresh docs build is the second commit.

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -207,7 +207,7 @@ paths:
         ^2c7cfa
         ```
         
-        ## Append Content Below a Heading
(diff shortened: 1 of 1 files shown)


coddingtonbear:
Thanks for the patch, @Marc-R2 !

PR #219, opened 2026-02-11
Add /tags/ endpoint

glnds:
## Summary
- Add `GET /tags/` endpoint exposing `MetadataCache.getTags()` — returns all vault tags with usage counts
- Add OpenAPI documentation for the endpoint
- Add tests (200 + 401 cases)

Supersedes #199 with all review feedback addressed (removed Postman collection, version bumps, CLAUDE.md, Makefile).

## Test plan
- [x] `npm test` — all 64 tests pass including new tags tests
- [x] `npm run build` — TypeScript compiles
- [x] `npm run build-docs` — OpenAPI regenerates cleanly

### DIFF
--- a/src/requestHandler.ts
+++ b/src/requestHandler.ts
@@ -1108,10 +1108,10 @@ export default class RequestHandler {
         }
       }
     }
-    const tags: Record<string, { count: number }> = {};
+    const tags: { name: string; count: number }[] = [];
     for (const [tag, count] of Object.entries(tagCounts)) {
       if (!tag) continue;
-      tags[tag] = { count };
+      tags.push({ name: tag, count });
     }
     res.json({ tags });
   }
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -1713,24 +1713,29 @@ paths:
             application/json:
               example:
                 tags:
-                  important:
-                    count: 1
-                  project:
-                    count: 3
-                  work/tasks:
-                    count: 2
+                  - count: 3
+                    name: "project"
+                  - count: 1
+                    name: "important"
+                  - count: 2
+                    name: "work"
+                  - count: 2
+                    name: "work/tasks"
               schema:
                 properties:
                   tags:
-                    additionalProperties:
+                    items:
                       properties:
                         count:
                           description: "Number of times this tag is used across the vault."
                           type: "number"
+                        name:
+                          description: "Tag name without the leading `#`."
+                          type: "string"
                       type: "object"
-                    type: "object"
+                    type: "array"
                 type: "object"
-          description: "A mapping of tag names to their metadata."
+          description: "A list of tags with their usage counts."
       summary: |
         Get a list of all tags with metadata.
       tags:
--- a/docs/src/openapi.jsonnet
+++ b/docs/src/openapi.jsonnet
@@ -390,17 +390,21 @@ std.manifestYamlDoc(
           description: 'Returns all tags found across all files in the vault, drawn from both inline (`#tag`) and frontmatter tag syntax. Each tag is returned without the `#` prefix. Hierarchical tags (e.g. `work/tasks`) also contribute a count to every parent prefix (e.g. `work`), mirroring how Obsidian displays tag counts in its sidebar.\n',
           responses: {
             '200': {
-              description: 'A mapping of tag names to their metadata.',
+              description: 'A list of tags with their usage counts.',
               content: {
                 'application/json': {
                   schema: {
                     type: 'object',
                     properties: {
                       tags: {
-                        type: 'object',
-                        additionalProperties: {
+                        type: 'array',
+                        items: {
                           type: 'object',
                           properties: {
+                            name: {
+                              type: 'string',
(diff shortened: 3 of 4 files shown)


src/requestHandler.ts
```diff
@@ -1089,6 +1089,15 @@ export default class RequestHandler {
     );
   }
 
+  async tagsGet(req: express.Request, res: express.Response): Promise<void> {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const allTags = (this.app.metadataCache as any).getTags();
```
coddingtonbear: I'm afraid this doesn't look like it's on the public interface:

https://github.com/obsidianmd/obsidian-api/blob/master/obsidian.d.ts#L4188

Is there conversation somewhere about that API becoming public in the future?

coddingtonbear (commented):
Thanks for updating this PR and pushing things forward a bit!  This PR looks great quality-wise now, but there are a couple things I'd love to talk through with you below first.  Just to be clear: there are some general questions and suggestions I'm making below, but this isn't all on you -- I'm glad to help out with the below API suggestions if you're in agreement about the rationale.

# API Suggestions

## Tag `#` prefix

The Obsidian API (and the one you've built atop it) returns tags with a `#` prefix, but I’d prefer we strip that for two reasons:

1. In the `/vault/` API's `PATCH` method, we also allow tags to be used for a particular purpose, and there we don't include the `#`.
2. In just a moment (in "Response Structure" below) you're going to hear me pitching for designing this API to allow for a future tag detail API, and if we were to have such an API, the tag name might end up in a URL, which would make that `#` symbol into a problem since `#` isn't allowed in the route part of a URL.

## Response Structure

I’d recommend we refactor this slightly to leave room for future flexibility around tag metadata.  Right now, there's just one endpoint and it returns something like this:

`/tags/`
```json
{
    "tags": {
        "sometag": 10,
        "another_tag": 348,
        "tag_3": 14
    }
}
```

You and I both know that that count next to each tag is apparently[^apparently] the count of instances of that tag being used, but:

1. That isn't an "obvious" thing -- somebody seeing those values would be equally valid in guessing that that integer is the tag's internal ID number as they would to guess it's the instance count.
2. The current shape (`Record<string, number>`) locks us into exactly one piece of information per tag. If we ever want to add additional metadata, we’d either need to introduce a breaking change or create parallel endpoints. Designing the response as an object now avoids that trap.
3. If we were to want to add additional information, what might make the most sense.  One choice is changing this mapping from a `Record<str, number>` to a `Record<str, Record<str, unknown>>`:

`/tags/`
```json
{
    "tags": {
        "sometag": {
            "count": 10,
            "last_used": "2025-12-12T10:01:00Z",
        },
        "another_tag": {
            "count": 348,
            "last_used": "2026-01-12T22:01:40Z",
        },
        "tag_3": {
            "count": 14,
            "last_used": "2025-12-10T01:04:56Z",
        }
    }
}
```

or, alternatively, maybe this should instead mirror the `/vault/ ` REST APIs[^consistency] and instead make this API return just an array:

`/tags/`
```json
{
    "tags": [
        "sometag",
        "another_tag",
        "tag_3"
    ]
}
```

& then we could (if we wanted to provide additional information) just add new endpoints for each tag:

`/tags/sometag/`
```
{
    "count": 10,
    "last_used": "2025-12-12T10:01:00Z",
}
```

I'm not totally sure which approach makes practical sense, but I know that consistency across endpoints dramatically reduces cognitive load for integrators -- especially in a REST surface that will continue to grow.  Cognitive load is the enemy of creativity.

[^apparently]: I haven't personally verified that that's what this number means, but I'm trusting that you have.
[^consistency]: Every engineer loves a consistent API. When multiple APIs follow similar patterns, it greatly reduces friction and makes it a lot more fun to integrate with something new.

# Undocumented API Use

This is, unfortunately, the big one!  I have a general concern about this relying on an undocumented API. I want to be clear: I totally trust you that right now this API exists and functions as you've used it here, but there's a non-obvious risk around use of an undocumented API in that an undocumented API could change, move, or disappear at any time without any notice.  At that point, there are lots of things that might happen ranging from breaking software dependent on the API to crashing or hanging Obsidian or at least the REST API.

I'm curious about how you came across this API at all -- did you find it documented somewhere or in use by some other plugin?

So, the bad news: I can’t accept a PR that depends on an undocumented API unless we can confirm it’s stable and supported. The good news, though: this is *exactly* the kind of thing [API extensions](https://github.com/coddingtonbear/obsidian-local-rest-api/wiki/Adding-your-own-API-Routes-via-an-Extension) are for.  I know adapting a PR to an API extension might seem a little daunting, but I'd be glad to help you out with putting that together if we couldn't find a way of guaranteeing this API going forward.

glnds:
Reworked to use only documented Obsidian APIs:

- `vault.getMarkdownFiles()`
- `metadataCache.getFileCache(file)`
- `getAllTags(cache)` from `obsidian`

No more `metadataCache.getTags()`. Parent tag roll-up is done manually (e.g. `#food/recipe/pasta` counts toward `#food/recipe` and `#food`), matching the previous behavior. Squashed into a single commit.

PR #239, opened 2026-04-23
Add opt-in last-active file fallback for /active/* endpoints

mgardestad:
## Summary
- Adds a `?use-last-active=true` query parameter to the `/active/*` endpoints. When no file is currently active, the request falls back to the most recently opened file instead of returning 404.
- Tracks the last-active file via a `file-open` workspace listener registered in `setupRouter`, with a `teardown()` cleanup wired into `onunload`.
- Responses set an `X-Obsidian-Note-Active` header (`true` for the live active file, `false` when served from the fallback) so callers can tell which they got.

## Test plan
- [ ] With a note open, GET `/active/` returns it and `X-Obsidian-Note-Active: true`.
- [ ] With no active file, GET `/active/` returns 404.
- [ ] With no active file, GET `/active/?use-last-active=true` returns the previously opened file with `X-Obsidian-Note-Active: false`.
- [ ] PUT/POST/PATCH/DELETE on `/active/` honor the same fallback behavior.
- [ ] Disabling and re-enabling the plugin does not leak the workspace event listener.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "obsidian-local-rest-api",
-  "version": "3.6.0",
+  "version": "3.6.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "obsidian-local-rest-api",
-      "version": "3.6.0",
+      "version": "3.6.1",
       "license": "MIT",
       "dependencies": {
         "@types/response-time": "^2.3.5",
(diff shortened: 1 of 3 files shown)


coddingtonbear:
Thanks for your patch, @mgardestad!  Sorry for the delay on this, but it took me a bit of time to think through whether I wanted to accept this patch into the core, and in the end I've landed on instead suggesting you implement this as an [API extension](https://github.com/coddingtonbear/obsidian-local-rest-api/wiki/Adding-your-own-API-Routes-via-an-Extension).  I know this isn't what you want to hear, but I hope you'll understand after I explain my rationale.

There are a few reasons I landed on this:
- Every option we add to the API makes the API and its behavior more difficult to reason about.  I really appreciate the care you took around this to make this backward-compatible by making the API opt-in, as well as communicating to consumers when the new functionality was used. The choices you made are definitely the right ones to make for preserving backward compatibility.  Unfortunately, though, complexity is also a concern to consider when adding to an API and the functionality added by any option needs to be worth the complexity it adds.
- This introduces a bit of semantic drift. The API endpoint is named "active" because it (when it was created) was intended to allow you to target whatever file you currently have active in the editor.  After this change, it might mean that, but it might also mean just that the file was recently opened.  That's not the worst thing in the world, but it's not ideal, either.
- There are some (minor) security concerns, though, too.  Imagine that some API consumer has intentionally granted some tool (or LLM) access to `/active/` with the belief that that means that that tool will only be able to access the file they currently have focused and open[^1] in their Obsidian instance.  After this change, that is no longer guaranteed to be true, and if that API consumer learns about this new parameter, it might be able to access more than the note owner intends.

I know this isn't what you want to hear, and I'm sorry about that!  Don't worry, though, refactoring this PR's work into an [API extension](https://github.com/coddingtonbear/obsidian-local-rest-api/wiki/Adding-your-own-API-Routes-via-an-Extension) is a lot easier than you might be guessing!  In the future, if you want to avoid investing effort in a patch that might not be accepted, I recommend starting a discussion on the Discussions tab before getting started; you can read more in [our contribution guidelines](https://github.com/coddingtonbear/obsidian-local-rest-api/blob/main/CONTRIBUTING.md).

[^1]: This might sound contrived, but I can say that I do use this kind of strategy a lot when giving LLMs access to tools.  For example, when using an LLM at work, I allow it to read from our issue tracker without me gatekeeping anything, but I configure what tools it has available to it such that it requires positive approval from me before trying to post a comment on that issue tracker.  In this context here, it's conceivable here that somebody might grant an LLM access to `/active/` without prompting, but want to approve accessing arbitrary notes.

mgardestad:
Thank you for getting back to me and also for pointing me to the [API extension](https://github.com/coddingtonbear/obsidian-local-rest-api/wiki/Adding-your-own-API-Routes-via-an-Extension). 
I was not aware of this plugin and it is a much cleaner solution.

PR #244, opened 2026-05-08
fix: pass app context to getAPI() so DataView is resolved when window.DataviewAPI is unset

gsjsbshsj9-a11y:
## Problem

`POST /search/` with `application/vnd.olrapi.dataview.dql+txt` content type currently returns **400** with body:

```
Cannot read properties of undefined (reading 'tryQuery')
```

### Reproduction

```sh
curl -k -X POST \
  -H 'Authorization: Bearer <key>' \
  -H 'Content-Type: application/vnd.olrapi.dataview.dql+txt' \
  --data 'TABLE file.name FROM ""' \
  https://127.0.0.1:27124/search/
```

## Root cause

In `src/requestHandler.ts` (`searchQueryPost`), the DataView API is resolved via `getAPI()` without an argument. Per [obsidian-dataview's `getAPI(app?: App)`](https://blacksmithgu.github.io/obsidian-dataview/api/intro/), when `app` is omitted the helper falls back to `window.DataviewAPI`, which can be `undefined` in REST/headless contexts. The handler then calls `.tryQuery(...)` on `undefined`, producing the user-facing 400 above.

## Fix

Pass `this.app` to `getAPI()` so it resolves the DataView API through the active plugin registry instead of relying on the global. The plugin registry is reliably populated when Dataview is enabled.

## Scope

- One-line change at the single call site (full diff in commit body).
- No behaviour change for callers that already had Dataview API resolved through the global.
- No changes to public API, types, or routes.

## Verification

- `npm run build` — passes (tsc + esbuild).
- `npm test` — 94/94 tests pass.
- After the fix, the curl above returns the expected JSON result for any valid DQL.

### DIFF
--- a/src/requestHandler.ts
+++ b/src/requestHandler.ts
@@ -1779,7 +1779,7 @@ export default class RequestHandler {
     req: express.Request,
     res: express.Response,
   ): Promise<void> {
-    const dataviewApi = getDataviewAPI();
+    const dataviewApi = getDataviewAPI(this.app);
 
     const handlers: Record<string, () => Promise<SearchJsonResponseItem[]>> = {
       [ContentTypes.dataviewDql]: async () => {


coddingtonbear (commented):
So, interestingly, I can confirm that Dataview queries are currently broken -- thanks for making me aware of that -- when I try to run them _before_ this patch, I see this:

```bash
ᐷ echo -n "TABLE file.link WHERE id = '39d259b3-3f80-4e5f-b5cd-27be34a3118d'" | http POST http://127.0.0.1:27123/search/ \ 
  "Authorization: Bearer $OBSIDIAN_LOCAL_REST_API_KEY" \
  "Content-Type: application/vnd.olrapi.dataview.dql+txt"
HTTP/1.1 400 Bad Request
Access-Control-Allow-Origin: *
Connection: keep-alive
Content-Length: 145
Content-Type: application/json; charset=utf-8
Date: Mon, 11 May 2026 00:16:03 GMT
ETag: W/"91-tDoK10eCqg6rJJv0lRS4BZ8NW6I"
Keep-Alive: timeout=5
X-Powered-By: Express
X-Response-Time: 2.836ms

{
    "errorCode": 40070,
    "message": "The query you provided could not be processed.\nCannot read properties of undefined (reading 'tryQuery')"
}
```

Unfortunately, though, it does not look like your patch here fixes the issue -- when I try what you have here, I see now this:

```bash
ᐷ echo -n "TABLE file.link WHERE id = '39d259b3-3f80-4e5f-b5cd-27be34a3118d'" | http POST http://127.0.0.1:27123/search/ \  
  "Authorization: Bearer $OBSIDIAN_LOCAL_REST_API_KEY" \
  "Content-Type: application/vnd.olrapi.dataview.dql+txt"
HTTP/1.1 400 Bad Request
Access-Control-Allow-Origin: *
Connection: keep-alive
Content-Length: 127
Content-Type: application/json; charset=utf-8
Date: Mon, 11 May 2026 00:16:54 GMT
ETag: W/"7f-soxe0SJF+aAItOV2afr2Udj0ey4"
Keep-Alive: timeout=5
X-Powered-By: Express
X-Response-Time: 3.673ms

{
    "errorCode": 40070,
    "message": "The query you provided could not be processed.\ndataviewApi.tryQuery is not a function"
}
```

Is there something I'm doing differently than you, here?

Unfortunately, unless you can provide a fix, I'm rather tempted to just remove Dataview support from this project entirely given that apparently nobody noticed that it has been broken for some time, and the Dataview project hasn't had active development in over a year.

coddingtonbear:
I've taken the step of removing support for DQL queries in 6011d23f19019cf773fb0036ffcd59626adca1fe .  I appreciate you taking the time to try to fix the query support, even if I have decided not to keep supporting that feature. Hearing from you that that query support was broken was helpful!

PR #252, opened 2026-05-20
Honor userIgnoreFilters in search and tag endpoints

fa1k3:
## Summary

This makes the vault-wide search and tag aggregation paths respect Obsidian's `userIgnoreFilters` setting.

Covered surfaces:

- `/tags/`
- `/search/simple/`
- `/search/` JSONLogic queries
- JSONLogic `links` / `backlinks` metadata derived from resolved links

The matcher keeps exact folder/file-prefix behavior for plain ignore entries and also supports glob-style ignore patterns via the existing `minimatch` dependency.

## Tests

- `npm run lint`
- `npm run typecheck`
- `npm run build`
- `npm test -- --runInBand`

`npm run test:integration -- --runInBand` was not run to completion because the integration suite requires `OBSIDIAN_API_KEY`.

### DIFF
--- a/src/vaultOperations.ts
+++ b/src/vaultOperations.ts
@@ -25,8 +25,19 @@ const WildcardRegexp = require("glob-to-regexp") as (pattern: string) => RegExp;
 const minimatch = require("minimatch") as (
   path: string,
   pattern: string,
-  options?: { dot?: boolean; nocase?: boolean },
+  options?: { dot?: boolean; nocase?: boolean; nonegate?: boolean; nocomment?: boolean },
 ) => boolean;
+
+const USER_IGNORE_GLOB_OPTIONS = {
+  dot: true,
+  nocase: false,
+  nonegate: true,
+  nocomment: true,
+};
+
+function hasGlobMagic(pattern: string): boolean {
+  return /[*?[\]{}]/.test(pattern) || /[!+@]\(/.test(pattern);
+}
 
 export class FileNotFoundError extends Error {}
 export class CommandNotFoundError extends Error {}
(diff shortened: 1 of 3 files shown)


src/vaultOperations.ts
```diff
@@ -59,6 +75,35 @@ export class VaultOperations {
     );
   }
 
+  private isUserIgnored(filePath: string): boolean {
+    const vault = this.app.vault as typeof this.app.vault & {
+      config?: { userIgnoreFilters?: string[] };
+    };
```
coddingtonbear, line 81: Am I correct in seeing that this is an undocumented API?  I don't see this documented here: https://github.com/obsidianmd/obsidian-api/blob/master/obsidian.d.ts; so I think so?

coddingtonbear (commented):
I'm sorry to say this, but I don't think this can land in core as-is for a couple of reasons:

1.  I would need to be very confident that `minimatch` and the implementation here exactly match Obsidian's own behavior. Otherwise, I worry we'd end up fielding bug reports where users are correctly pointing out that this API behaves differently from Obsidian itself.
2.  Ideally, we'd rely on the Obsidian API itself to handle this filtering. Is that possible here? My assumption is that it may not be, since this appears to depend on an undocumented API.

Unfortunately, either of those would be a blocker for including this in core.

I know that's probably not the answer you were hoping for, and I appreciate the work you put into this. You can, of course, still implement a similar feature as an [API extension](https://github.com/coddingtonbear/obsidian-local-rest-api/wiki/Adding-your-own-API-Routes-via-an-Extension).

For future changes of this size, please start with a Discussion in the "Discussions" tab before investing too much implementation time. That gives us a chance to talk through whether the feature is likely to be accepted into core first. The [contribution guidelines](https://github.com/coddingtonbear/obsidian-local-rest-api/blob/main/CONTRIBUTING.md) have more information on that process.

PR #253, opened 2026-05-20
Configure server keep-alive timeouts

fa1k3:
## Summary

- set explicit keep-alive and headers timeouts on both secure and insecure HTTP servers
- keep `headersTimeout` slightly above `keepAliveTimeout`, matching Node's recommended relationship
- add unit coverage for the timeout helper

## Why

The default Node keep-alive timeout is short. Explicitly configuring these values avoids avoidable stale-socket/RST races for clients that reuse connections.

## Tests

- `npm test -- src/serverTimeouts.test.ts --runInBand`
- `npm run typecheck`
- `npm run build`
- `npm test -- --runInBand`
- `npm run lint` (passes with one existing warning in `src/integration/patch.test.ts`)

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -20,6 +20,7 @@ import {
 import LocalRestApiPublicApi, { ApiVersionUnsupportedError } from "./api";
 export { ApiVersionUnsupportedError } from "./api";
 import { PluginManifest } from "obsidian";
+import { configureHttpServerTimeouts } from "./serverTimeouts";
 
 export default class LocalRestApi extends Plugin {
   settings: LocalRestApiSettings;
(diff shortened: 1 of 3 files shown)


coddingtonbear (approved):

coddingtonbear:
I've merged this because it's harmless, but I am curious, though, @fa1k3 -- what's the context for needing those particular values set?  I'd _imagine_ that they'd only be things you'd encounter if you were running this API behind a reverse proxy.  Can you help me understand your use case?

fa1k3:
Thanks for merging this!

Context is a local Obsidian automation setup talking directly to Local REST API over localhost, no reverse proxy. I was seeing occasional connection reuse / stale socket issues with Node clients, so this was meant to make the server timeout behavior explicit.

The exact values are not special. Happy to adjust them or add a short comment if you prefer.

PR #254, opened 2026-05-20
mcp url typo fix

suleyman-shb:
Had trouble with urls, it was working but also timing out.
Thanks


coddingtonbear:
I appreciate the help, @suleyman-shb, but it looks like this PR includes a lot of unrelated changes.  Can you post this again with just the specific change you made?

coddingtonbear:
I looked at your commits individually and saw that your only intended change was probably just https://github.com/coddingtonbear/obsidian-local-rest-api/pull/254/commits/6767e3090104498fbe120288b01e13cd946e5dc1 -- I know that this kind of thing is a little unfamiliar to a lot of people, but in the future, it's a good idea to always create a branch before posting a PR so that unrelated changes you make do not end up automatically appearing in pull requests you earlier posted.

About the specific change, though:  I'd be more than a bit surprised if there were a functional difference between `https://127.0.0.1:${this.plugin.settings.port}/mcp/` and `https://127.0.0.1:${this.plugin.settings.port}/mcp` -- I don't see a difference in my environment between the two, and my understanding of Express (the web server used by this project) is that they are both as the same "normalized" path, but all of the other documentation uses `/mcp/` and it's probably better for me to be consistent anyway, so I'm going to make a similar change via 79fbdf53c15712410b628225de4c765901522e83.

That all being said, I'd _suspect_ that you swapping `/mcp` for `/mcp/` fixing your problem was just a red herring!  Good luck, though, and thanks!

suleyman-shb:
Wow, that was fast, thanks. Yes, there were other commits included, sorry about that. If it makes no difference then sorry for the time spent, thanks for the development, cheers.

Issue #251, opened 2026-05-19
MCP server: SSE stream disconnects immediately on every tools/call (Streamable HTTP, mcp-remote bridge, Windows)


joehyn21-sys:
### Summary

The built-in MCP server accepts connections and successfully completes
`initialize`, `tools/list`, and `resources/list`, but **every `tools/call`
fails**. The SSE stream is terminated almost immediately after the call is
issued, and the client request eventually times out. Reconnecting (toggling
the MCP server off/on, restarting the client) does not help — the new stream
dies the same way.

### Environment

- Plugin: Local REST API & MCP Server, version 4.0.2
- Obsidian: 1.12.7
- OS: Windows
- MCP client: Claude Desktop (MSIX build)
- Bridge: mcp-remote 0.1.37 / 0.1.38 (via `npx.cmd`)
- Node.js: v24.15.0
- Endpoint used: `http://127.0.0.1:27123/mcp/` (plain HTTP endpoint,
  "Enable non-encrypted (HTTP) server" turned on)
- Auth: `Authorization: Bearer <api-key>` header (authentication itself
  works — see notes)

### What works

- `GET http://127.0.0.1:27124/` returns `{"status":"OK", ...}` (server is up)
- MCP `initialize` succeeds
- `tools/list` returns the full tool list correctly
- `resources/list` returns the openapi-spec resource correctly

So the handshake, authentication, and capability negotiation are all fine.

### What fails

Any actual `tools/call` (e.g. `vault_list` with `{"path":""}` or a folder
path) never returns. The bridge logs show the stream being terminated
repeatedly:

Error from remote server: Error: SSE stream disconnected: TypeError: terminated
at processStream (.../mcp-remote/dist/chunk-...js)

and, when a tools/call is in flight:
[TypeError: fetch failed] {
[cause]: HeadersTimeoutError: Headers Timeout Error
code: 'UND_ERR_HEADERS_TIMEOUT'
}

The client then cancels with:
notifications/cancelled  reason: "McpError: MCP error -32001: Request timed out"

This repeats on every reconnect. The stream does not survive long enough to
complete a single tool call.

### Steps to reproduce

1. Enable the plugin, turn on the non-encrypted HTTP server (port 27123).
2. Configure Claude Desktop via mcp-remote bridge:
```json
   {
     "mcpServers": {
       "obsidian": {
         "command": "npx.cmd",
         "args": [
           "-y",
           "mcp-remote",
[...]

PR #267, opened 2026-05-28
Fix MCP server hanging on all but the most-recently-connected session

psalkowski:
## Summary

The built-in MCP server shares **one `McpServer` instance across every session's
transport**. The MCP SDK's `Server.connect()` binds a single `_transport`, so each
new session silently repoints that reference to its own transport. From then on,
any request arriving on an **older** session has its response routed to the
**newest** session's socket — the older request never gets a reply and hangs until
the client times out (~60s), and the misrouted reply lands on the wrong session.

In practice: as soon as a second MCP client (or a reconnect, or even a second
`initialize`) connects, every previously-connected session stops working — while
still showing as "connected". This matches the reports in
#266 and #263 (timeouts/disconnects that "reconnect usually fixes", multiple
clients, intermittent).

## Root cause

`src/mcpHandler.ts` builds one server in the constructor and connects every new
session's transport to it:

```ts
await this.mcpServer.connect(transport); // shared server — rebinds _transport
```

`initialize` / `tools/list` / `resources/list` are simple request→response and keep
working (which is why the connection still looks healthy), but `tools/call` relies
on the per-session response routing that the shared `_transport` breaks.

## Fix

Give every session its own `McpServer`. Tool/resource registrations are captured
once as reusable specs at construction time and replayed onto a fresh server for
each new transport (`buildServer()`), kept paired 1:1 in the session map. The
public `registerTool()` extension API is preserved.

## Steps to reproduce

With the plugin running and the non-encrypted (HTTP) server enabled on port 27123:

```bash
TOKEN=<API_KEY>
H=(-H "Authorization: Bearer $TOKEN" \
   -H "Accept: application/json, text/event-stream" \
   -H "Content-Type: application/json" \
   -H "MCP-Protocol-Version: 2025-03-26")

# Session A — initialize, capture its session id
A=$(curl -sS -i -X POST http://127.0.0.1:27123/mcp/ "${H[@]}" \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-03-26","capabilities":{},"clientInfo":{"name":"A","version":"1"}}}')
SIDA=$(echo "$A" | grep -i '^mcp-session-id:' | awk '{print $2}' | tr -d '\r')
curl -sS -X POST http://127.0.0.1:27123/mcp/ "${H[@]}" -H "Mcp-Session-Id: $SIDA" \
  -d '{"jsonrpc":"2.0","method":"notifications/initialized"}' >/dev/null

# Session B — initialize (steals the shared server's _transport)
B=$(curl -sS -i -X POST http://127.0.0.1:27123/mcp/ "${H[@]}" \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-03-26","capabilities":{},"clientInfo":{"name":"B","version":"1"}}}')
SIDB=$(echo "$B" | grep -i '^mcp-session-id:' | awk '{print $2}' | tr -d '\r')
curl -sS -X POST http://127.0.0.1:27123/mcp/ "${H[@]}" -H "Mcp-Session-Id: $SIDB" \
  -d '{"jsonrpc":"2.0","method":"notifications/initialized"}' >/dev/null

# Tool call on the OLDER session A — hangs; on the NEWER session B — instant
curl -sS -o /dev/null -w "A: %{size_download} bytes in %{time_total}s\n" --max-time 12 \
  -X POST http://127.0.0.1:27123/mcp/ "${H[@]}" -H "Mcp-Session-Id: $SIDA" \
  -d '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"vault_list","arguments":{"path":""}}}'
curl -sS -o /dev/null -w "B: %{size_download} bytes in %{time_total}s\n" --max-time 12 \
  -X POST http://127.0.0.1:27123/mcp/ "${H[@]}" -H "Mcp-Session-Id: $SIDB" \
  -d '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"vault_list","arguments":{"path":""}}}'
```

**Before the fix:**

```
A: 0 bytes in 12.006s     # older session hangs
B: 158 bytes in 0.004s    # newer session works
```

Packet capture on loopback confirms the older session's `tools/call` is accepted
at the TCP layer (POST body ACKed) and then gets **zero bytes** of HTTP response
until the client RSTs ~60s later, while a fresh session runs the identical tool in
single-digit milliseconds.

## Evidence the fix works

**After the fix**, the same reproduction returns results on **both** sessions:

```
A: 158 bytes in 0.007s
B: 158 bytes in 0.002s
```

Concurrency check — 10 sessions opened in sequence, then a `tools/call` fired on
every session (including the very first) plus the oldest session hammered 5 more
times:

```
opened 10 sessions
wave1 (all 10 concurrent): 10/10 ok
wave2 (oldest session x5): 5/5 ok
RESULT: 15/15 PASS
```

## Tests

`src/mcpHandler.test.ts` updated for per-session registration (tools/resources are
now registered when a session is built, not at construction). Full suite green:

- `npm test` — 187 passed
- `npm run typecheck` — clean
- `npm run lint` — clean
- `npm run build` — ok

## Related

- Discussion #266
- Discussion #263 / #251 (same shape: handshake + `tools/list` succeed, `tools/call` hangs)

### DIFF
--- a/src/integration/mcp.test.ts
+++ b/src/integration/mcp.test.ts
@@ -625,3 +625,32 @@ describe("periodic_note_get_path tool", () => {
     expect(typeof body.path).toBe("string");
   });
 });
+
+// ---------------------------------------------------------------------------
+// Multi-session routing (regression for shared-McpServer bug)
+// ---------------------------------------------------------------------------
+
+describe("multi-session routing", () => {
+  test("first session remains functional after a second session connects", async () => {
+    const clientA = makeClient();
+    const clientB = makeClient();
+    try {
+      await clientA.connect(makeTransport());
+      // Connecting clientB previously overwrote the shared McpServer's internal
+      // _transport reference, causing clientA's subsequent tool calls to hang
+      // indefinitely as their responses were routed to clientB's transport.
+      await clientB.connect(makeTransport());
+
+      const resultA = await clientA.callTool({ name: "vault_list", arguments: {} });
+      expect(resultA.isError).toBeFalsy();
+      expect(Array.isArray(jsonOf<{ files: string[] }>(resultA).files)).toBe(true);
+
+      const resultB = await clientB.callTool({ name: "vault_list", arguments: {} });
+      expect(resultB.isError).toBeFalsy();
+      expect(Array.isArray(jsonOf<{ files: string[] }>(resultB).files)).toBe(true);
+    } finally {
+      await clientA.close().catch((_e: unknown): void => {});
+      await clientB.close().catch((_e: unknown): void => {});
+    }
+  });
+});
--- a/src/mcpHandler.test.ts
+++ b/src/mcpHandler.test.ts
@@ -136,11 +136,22 @@ describe("McpHandler", () => {
    
   let ops: any;
 
-  beforeEach(() => {
+  // Each session now gets its own McpServer, so tool/resource registration runs when
+  // a session is built (not at construction). Trigger one build so the McpServer mock
+  // captures the registrations that getToolCallback() reads.
+  async function buildSession(mcp: McpHandler): Promise<void> {
+    // @ts-ignore: partial mock
+    await mcp.handleRequest(
+      { headers: {}, body: { jsonrpc: "2.0", id: 0, method: "initialize" } },
+      { status: jest.fn().mockReturnThis(), json: jest.fn() },
+    );
+  }
+
+  beforeEach(async () => {
     jest.clearAllMocks();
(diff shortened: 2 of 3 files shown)


coddingtonbear:
Good find!  Thanks for posting a fix!  I think I can sort through the next steps to get this across the finish line tonight.

PR #268, opened 2026-05-28
Fix MCP vault_patch storing application/json content as a quoted string

psalkowski:
## Summary

The built-in MCP server's `vault_patch` tool stores `application/json` content **exactly as it receives it**. MCP clients (LLM agents) routinely pass that content as a **JSON-encoded string** (e.g. `"[\"a\",\"b\"]"`, `"null"`) rather than as a native JSON value, so the string is handed straight to `markdown-patch`, whose `yaml.stringify` then serialises *the string itself*. The results:

- `["a","b"]` → `related: '["a","b"]'` — a quoted scalar, **not** a YAML sequence (frontmatter lists are effectively impossible via MCP)
- `null` → `tags: "null"` — the literal string `"null"`
- `"[[link]]"` → `plan: '"[[link]]"'` — double-encoded
- `append`/`prepend` concatenate the serialised strings instead of merging lists

The REST `PATCH /vault/{path}` endpoint is **not** affected: the global `express.json` middleware parses the request body into a native value before it reaches the patch logic. Only the MCP path lacked that step, so MCP and REST diverged.

## Root cause

`src/mcpHandler.ts`'s `vault_patch` callback passes `content` (typed `z.unknown()`) straight through to `patchFileSection` → `applyPatch`. For a frontmatter target, `markdown-patch` ignores `contentType` and simply `yaml.stringify`s whatever JS value it is handed — so a native array becomes a sequence, but a *string* becomes a quoted scalar. Because nothing on the MCP path parses the JSON, a stringified payload round-trips as a quoted string. (`vault_read` re-parses YAML, so the corruption is only visible in the raw file bytes.)

## Fix

Mirror what `express.json` already does on the REST path: in the MCP `vault_patch` handler, when `contentType === "application/json"` and `content` is a string, `JSON.parse` it into a native value before patching, and reject malformed JSON (matching express's 400-on-bad-JSON). Native (non-string) content passes through unchanged. The guard is gated on `Content-Type` for all target types, exactly like the global `express.json` middleware.

```ts
let parsedContent = content;
if ((contentType ?? "text/markdown") === "application/json" && typeof content === "string") {
  try {
    parsedContent = JSON.parse(content);
  } catch (err) {
    throw new Error(`Invalid application/json content: ${err instanceof Error ? err.message : String(err)}`);
  }
}
```

## Reproduce (MCP)

Call the `vault_patch` MCP tool against a note that has frontmatter:

- `targetType: frontmatter`, `target: related`, `operation: replace`, `contentType: application/json`, `content: ["a","b"]`

**Before:** the field is written as `related: '["a","b"]'` (a quoted string). `content: null` is written as `tags: "null"`.
**After:** `related:` becomes a real sequence (`- a` / `- b`), and `null` becomes native `null`.

## Tests

- `src/mcpHandler.test.ts` — stringified array → native array; `"null"` → `null`; JSON-encoded scalar without double-encoding; malformed JSON throws and does not patch; `text/markdown` content is left unparsed.
- `src/integration/mcp.test.ts` — end-to-end: an `application/json` array is stored as a real list; malformed `application/json` is rejected.

Body/prose content is unaffected — it uses `text/markdown`, which the parse step never touches.

### DIFF
--- a/src/mcpHandler.ts
+++ b/src/mcpHandler.ts
@@ -331,7 +331,7 @@ export class McpHandler {
           .describe("How to apply the content: replace the section, prepend before it, or append after it"),
         content: z.unknown().describe("Content to apply. For contentType 'text/markdown' pass a string. For contentType 'application/json' you may pass a native JSON value (number, boolean, array, object) and it will be serialised automatically."),
         contentType: z
-          .string()
+          .enum(["text/markdown", "application/json"])
           .optional()
           .describe(
             "MIME type of content. 'text/markdown' (default) or 'application/json'. " +
@@ -382,7 +382,7 @@ export class McpHandler {
         target: string;
         operation: PatchOperation;
         content: unknown;
-        contentType?: string;
+        contentType?: "text/markdown" | "application/json";
         createTargetIfMissing?: boolean;
         trimTargetWhitespace?: boolean;
         rejectIfContentPreexists?: boolean;
--- a/src/mcpHandler.ts
+++ b/src/mcpHandler.ts
@@ -329,7 +329,7 @@ export class McpHandler {
         operation: z
           .enum(["replace", "prepend", "append"])
           .describe("How to apply the content: replace the section, prepend before it, or append after it"),
-        content: z.unknown().describe("Content to apply. For contentType 'text/markdown' pass a string. For contentType 'application/json' you may pass a native JSON value (number, boolean, array, object) and it will be serialised automatically."),
+        content: z.string().describe("Content to apply. For contentType 'text/markdown' pass markdown text. For contentType 'application/json' pass a JSON-encoded string (e.g. '[\"row\",\"cells\"]' for a table row, or '42' for a number)."),
         contentType: z
           .nativeEnum(ContentType)
           .optional()
@@ -381,7 +381,7 @@ export class McpHandler {
         targetType: PatchTargetType;
         target: string;
         operation: PatchOperation;
-        content: unknown;
+        content: string;
         contentType?: ContentType;
         createTargetIfMissing?: boolean;
         trimTargetWhitespace?: boolean;
@@ -390,15 +390,11 @@ export class McpHandler {
         targetScope?: "content" | "marker" | "markerAndContent";
       }) => {
         try {
-          // The REST path has express.json middleware that parses application/json
-          // bodies into native values before they arrive here; the MCP transport
-          // does not, so a string payload must be parsed explicitly.
+          // MCP transport delivers all parameters as strings; parse JSON content
+          // here so downstream code receives a native value, not a serialized string.
           const resolvedContentType = contentType ?? ContentType.text;
-          let parsedContent = content;
-          if (
-            resolvedContentType === ContentType.json &&
-            typeof content === "string"
-          ) {
+          let parsedContent: unknown = content;
+          if (resolvedContentType === ContentType.json) {
             try {
               parsedContent = JSON.parse(content);
             } catch (err) {
--- a/src/mcpHandler.ts
+++ b/src/mcpHandler.ts
@@ -395,9 +395,10 @@ export class McpHandler {
           // routinely pass it as a JSON-encoded string, so parse strings here
           // into a native value before patching (otherwise yaml.stringify would
           // store the raw string). Malformed JSON is rejected, as express does.
+          const resolvedContentType = contentType ?? "text/markdown";
           let parsedContent = content;
           if (
-            (contentType ?? "text/markdown") === "application/json" &&
+            resolvedContentType === "application/json" &&
             typeof content === "string"
           ) {
             try {
@@ -414,7 +415,7 @@ export class McpHandler {
             target,
             operation,
             parsedContent,
-            contentType ?? "text/markdown",
+            resolvedContentType,
             { createTargetIfMissing, trimTargetWhitespace, rejectIfContentPreexists, targetDelimiter, targetScope },
           );
         } catch (e) {
--- a/src/mcpHandler.ts
+++ b/src/mcpHandler.ts
@@ -390,12 +390,30 @@ export class McpHandler {
         targetScope?: "content" | "marker" | "markerAndContent";
       }) => {
         try {
+          // Mirror express.json on the REST path: when the caller declares
+          // application/json, the content is interpreted as JSON. MCP clients
+          // routinely pass it as a JSON-encoded string, so parse strings here
+          // into a native value before patching (otherwise yaml.stringify would
+          // store the raw string). Malformed JSON is rejected, as express does.
+          let parsedContent = content;
+          if (
+            (contentType ?? "text/markdown") === "application/json" &&
+            typeof content === "string"
+          ) {
+            try {
+              parsedContent = JSON.parse(content);
+            } catch (err) {
+              throw new Error(
+                `Invalid application/json content: ${err instanceof Error ? err.message : String(err)}`,
+              );
+            }
+          }
           await this.ops.patchFileSection(
             path,
             targetType,
             target,
             operation,
-            content,
+            parsedContent,
             contentType ?? "text/markdown",
             { createTargetIfMissing, trimTargetWhitespace, rejectIfContentPreexists, targetDelimiter, targetScope },
           );
--- a/src/mcpHandler.ts
+++ b/src/mcpHandler.ts
@@ -348,7 +348,7 @@ export class McpHandler {
         rejectIfContentPreexists: z
           .boolean()
           .optional()
-          .describe("If true, fail the patch when the content already appears in the target section (default: false). Use to make append/prepend operations idempotent on retry."),
+          .describe("If true, fail the patch when the content already appears in the target section (default: false). Use to make append/prepend operations idempotent on retry. Only applies to string ('text/markdown') content; it has no effect on 'application/json' content (e.g. table rows), which is not a string and so is never compared."),
         targetDelimiter: z
           .string()
           .optional()
--- a/src/mcpHandler.ts
+++ b/src/mcpHandler.ts
@@ -8,7 +8,7 @@ import express from "express";
 import { TFile } from "obsidian";
 
 import { VaultOperations } from "./vaultOperations";
-import { PatchFailed, PatchOperation, PatchTargetType } from "markdown-patch";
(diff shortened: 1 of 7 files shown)


src/mcpHandler.ts
```diff
@@ -333,12 +333,30 @@ export class McpHandler {
...
       }) => {
         try {
+          // Mirror express.json on the REST path: when the caller declares
+          // application/json, the content is interpreted as JSON. MCP clients
+          // routinely pass it as a JSON-encoded string, so parse strings here
+          // into a native value before patching (otherwise yaml.stringify would
+          // store the raw string). Malformed JSON is rejected, as express does.
+          let parsedContent = content;
+          if (
+            (contentType ?? "text/markdown") === "application/json" &&
```
coddingtonbear: **Bug (high): arbitrary `contentType` strings reach `applyPatch` unvalidated and can silently corrupt the vault file.**

The REST path rejects unknown MIME types at `isContentType(contentType)` before calling `patchFileSection`. The MCP path has no equivalent guard — `contentType` is declared as `z.string().optional()` with no enum constraint, so any string passes through.

`markdown-patch`'s `applyPatch` switches on `ContentType.text` / `ContentType.json` with no `default` branch. Any unrecognised MIME type causes `replace`/`append`/`prepend` to return `undefined`, which is then written back to the vault via `vault.adapter.write(filePath, undefined)`, corrupting the file silently.

**Failure scenario:** MCP client sends `contentType: "text/plain"` → this `"application/json"` guard skips → `patchFileSection` called with `"text/plain"` → `applyPatch` returns `undefined` → file corrupted.

**Suggested fix:** constrain the schema to the allowed enum, matching the REST path:
```ts
contentType: z
  .enum(["text/markdown", "application/json"])
  .optional()
```

src/mcpHandler.ts
```diff
@@ -333,12 +333,30 @@ export class McpHandler {
...
+          }
           await this.ops.patchFileSection(
             path,
             targetType,
             target,
             operation,
-            content,
+            parsedContent,
             contentType ?? "text/markdown",
             { createTargetIfMissing, trimTargetWhitespace, rejectIfContentPreexists, targetDelimiter, targetScope },
```
coddingtonbear, line 413: **Bug (medium): `rejectIfContentPreexists` is silently bypassed for any `application/json` content that parses to a non-string native value.**

After this change, `parsedContent` can be a number, boolean, array, or object. `markdown-patch`'s preexistence check begins with `typeof instruction.content === 'string'` and short-circuits immediately to `false` for any non-string — the idempotency guard is never evaluated.

**Failure scenario:** Agent calls `vault_patch` with `content: '42'`, `contentType: 'application/json'`, `rejectIfContentPreexists: true`, `operation: 'replace'` on a heading target. The string `'42'` is parsed to the number `42`. `markdown-patch` sees `typeof 42 === 'string'` → `false` → skips the preexistence check → patch always proceeds regardless of current document state, silently duplicating content on retries.

Note: `rejectIfContentPreexists` applies only to heading/block targets in `markdown-patch`; the frontmatter path has no such guard. The bug is present whenever a heading or block target is used with `application/json` content that resolves to a non-string.

src/mcpHandler.ts
```diff
@@ -333,12 +333,30 @@ export class McpHandler {
...
+            }
+          }
           await this.ops.patchFileSection(
             path,
             targetType,
             target,
             operation,
-            content,
+            parsedContent,
             contentType ?? "text/markdown",
```
coddingtonbear: **Simplification: `contentType ?? 'text/markdown'` is evaluated twice with no intermediate variable.**

Once on line 343 (the guard) and again on line 360 (the `patchFileSection` argument). If someone edits the default in one place without the other, the guard and the call diverge silently — the parse branch fires on one default while `patchFileSection` receives a different `contentType`.

```ts
const resolvedContentType = contentType ?? "text/markdown";
if (resolvedContentType === "application/json" && typeof content === "string") {
  // ...
}
await this.ops.patchFileSection(
  // ...
  resolvedContentType,
  // ...
);
```

src/mcpHandler.ts
```diff
@@ -333,12 +333,30 @@ export class McpHandler {
         targetScope?: "content" | "marker" | "markerAndContent";
       }) => {
         try {
+          // Mirror express.json on the REST path: when the caller declares
+          // application/json, the content is interpreted as JSON. MCP clients
+          // routinely pass it as a JSON-encoded string, so parse strings here
+          // into a native value before patching (otherwise yaml.stringify would
+          // store the raw string). Malformed JSON is rejected, as express does.
+          let parsedContent = content;
```
coddingtonbear: **Altitude: JSON-string coercion placed in the tool handler rather than `patchFileSection`.**

`patchFileSection` already accepts `content: unknown` and `contentType: string` — it is the natural place to normalise a JSON-encoded string into a native value, because that coercion is a property of how `contentType: 'application/json'` is supposed to behave, not a property of the MCP transport.

As written, any future caller of `patchFileSection` that passes a string `content` with `contentType === 'application/json'` (e.g., a new MCP tool, an internal utility) would silently store the raw string in the vault unless it independently reimplements this parse block. The REST path avoids the problem only because `express.json()` middleware handles it before the route handler — not because of anything in `patchFileSection` itself.

src/mcpHandler.ts
psalkowski: Fixed in [fa23717](https://github.com/coddingtonbear/obsidian-local-rest-api/pull/268/commits/fa237179902387ea540e11266ccc9e1c61e7cf98). The MCP `contentType` schema is now `z.enum(["text/markdown", "application/json"]).optional()`, matching the `isContentType` guard on the REST path, so an unrecognised MIME type is rejected at the tool boundary before it can reach `applyPatch`. As a backstop for any future caller, [61a43b4](https://github.com/coddingtonbear/obsidian-local-rest-api/pull/268/commits/61a43b4d2c411bc46dee0b4e73d1b2233dd67b85) also makes `patchFileSection` throw rather than write a non-string `applyPatch` result.

src/mcpHandler.ts
psalkowski: You're right that there's a real gap here, and it's this change that surfaces it. Tracing it through:

Before this PR, an MCP `application/json` table append arrived as a raw JSON *string* (`'[["a","b"]]'`). `appendTable` rejects anything that isn't `isStringArrayArray`, so it threw `ContentTypeInvalid` — table appends from MCP didn't actually work. This PR parses that string into a real array, so they work now. And because they now work, `rejectIfContentPreexists` becomes reachable for that content — but it's gated on `typeof instruction.content === 'string'` in `applyPatch`, so for the array it's silently skipped and a retry can duplicate the row. So the idempotency guard genuinely doesn't cover the case this PR enables.

One detail on the repro: it uses `operation: 'replace'`, but `applyPatch` only evaluates the preexistence check for non-`replace` operations, so that exact example wouldn't reach the gate regardless. The substantive case is append/prepend, which is where the duplication can actually happen.

The dedup itself can't be done correctly from this layer without reimplementing `markdown-patch`'s table internals — the right place to honour `rejectIfContentPreexists` for table rows is inside its `appendTable`/`prependTable`. I'm happy to file that upstream. For this PR, I've made the contract honest so a caller doesn't rely on idempotency that isn't there: [870be57](https://github.com/coddingtonbear/obsidian-local-rest-api/pull/268/commits/870be57139446c56947d7a87862070051cf695ec) documents in both the MCP tool description and the OpenAPI `Reject-If-Content-Preexists` parameter that the flag applies only to string (`text/markdown`) content. If you'd rather hold the table-append fix until `markdown-patch` can dedup rows, I can scope that change out instead.

src/mcpHandler.ts
psalkowski: Fixed in [a0dd618](https://github.com/coddingtonbear/obsidian-local-rest-api/pull/268/commits/a0dd6183de93575240bb0bae6f73d38c70eec903). The default is hoisted into a single `resolvedContentType` variable used by both the JSON-parse guard and the `patchFileSection` argument, so they can't diverge.

src/mcpHandler.ts
psalkowski: Agreed that `patchFileSection` should protect itself, but moving the JSON parse into it would break the REST path. REST mounts `express.json({ strict: false })`, so an `application/json` body arrives already decoded — including primitives: `"[[../plans/foo]]"` becomes the JS string `[[../plans/foo]]`. Re-parsing that in `patchFileSection` (`JSON.parse("[[../plans/foo]]")`) would throw, regressing the exact frontmatter-string case this PR adds. The parse compensates for the MCP transport having no `express.json` in front of it, so it belongs at the MCP boundary.

The future-caller corruption risk I handled at the `patchFileSection` altitude in [61a43b4](https://github.com/coddingtonbear/obsidian-local-rest-api/pull/268/commits/61a43b4d2c411bc46dee0b4e73d1b2233dd67b85): it now throws instead of writing a non-string `applyPatch` result, closing the path for every caller without assuming a wire format.

coddingtonbear:
Forgive me for the delay on this one @psalkowski -- I've had really limited free time over the last couple weeks, but I expect to find some time to look through this more thoroughly this weekend.

coddingtonbear:
Sorry again, @psalkowski; I've really been struggling to find the free time to look into this, but I promise, this is still on my radar and I intend to find some time to look through this in the next week or so :crossed_fingers: .

coddingtonbear (approved):
Thanks for the patch!

I made a few additional changes while reviewing:

- Replaced the hardcoded `"text/markdown"` / `"application/json"` string literals with the `ContentType` enum from `markdown-patch`, so there's one canonical source of truth for those values.
- Removed the guard that refused to write when `applyPatch` returned a non-string -- that path is unreachable by contract, and a bug in `markdown-patch` should be fixed there rather than papered over here. (Filed [markdown-patch#14](https://github.com/coddingtonbear/markdown-patch/issues/14).)
- Removed the caveat from the `rejectIfContentPreexists` docs noting it only works for string content — that's a bug in `markdown-patch`, not an intended limitation, so we shouldn't document it as expected behavior. (Filed [markdown-patch#13](https://github.com/coddingtonbear/markdown-patch/issues/13).)
- Tightened the `content` Zod schema from `z.unknown()` to `z.string()`, which is what MCP transport actually delivers regardless of schema. This also removes the `typeof` guard from the parse branch since the type is now enforced by Zod.

This'll go out with the next release!

psalkowski:
thanks @coddingtonbear!

PR #287, opened 2026-07-05
Fix: frontmatter append/prepend type-mismatch returns 500 instead of 400

### LINKED ISSUE #282: Frontmatter `append` with a type-mismatched value returns HTTP 500 instead of a 4xx client error
(I'm using MCP and Claude has identified an issue that it stumbled on, written up below)

### Summary

When `PATCH`-ing a frontmatter field with `Operation: append`, supplying a value whose type doesn't match the existing field (e.g. appending a scalar **string** to a field that is a **list**) returns **HTTP 500 Internal Server Error** with `errorCode 50000`. This is a client/usage error, not a server fault, and should return a `400 Bad Request` with a clear message.

The misleading `500` also propagates through the bundled MCP `vault_patch` tool, where it surfaces as a bare `Cannot merge objects of different types or unsupported types: object and string` — hard for a caller to interpret as "you passed the wrong shape."

### Steps to reproduce

Note with a list-valued frontmatter field:

```yaml
---
tags:
  - theme/orig
---
```

Append a JSON scalar string to it:

```bash
curl -sk -X PATCH 'http://127.0.0.1:27123/vault/Note.md' \
  -H 'Operation: append' \
  -H 'Target-Type: frontmatter' \
  -H 'Target: tags' \
  -H 'Content-Type: application/json' \
  --data '"theme/added"'
```

Response:

```json
{
  "message": "Internal Server Error\nCannot merge objects of different types or unsupported types: object and string",
  "errorCode": 50000
}
```
HTTP status: `500`

(For contrast, appending a JSON **array** `["theme/added"]` works and returns `200`.)

### Expected behavior

A type mismatch between the appended value and the existing frontmatter value is a predictable client error. It should return `400 Bad Request` (a `4xxx` `errorCode`) with a message that names the mismatch and the expected shape — e.g. *"cannot append a string to a list-valued field; provide an array"* — rather than a generic `500 Internal Server Error`.

### Impact

- Callers (and wrapper layers like the MCP `vault_patch` tool) can't distinguish "I sent the wrong body shape" from an actual server failure.
- `500`s are typically treated as non-retryable server faults, masking what is really a fixable request.

joshazmy:
## Problem

Closes #282.

`PATCH` with `Target-Type: frontmatter` and `Operation: append`/`prepend` returns an
unhandled **500** (instead of a clean 400) when the value being appended is incompatible
with the existing field's type — e.g. appending a string onto a field that already holds
a list.

## Root cause

`mergeFrontmatterValue()` throws a bare `Error` in this case (append-able type, but
incompatible with the existing field type). That error was never caught at the shared
REST/MCP call site in `VaultOperations.patchFileSection`, so it propagated up and was
reported as a generic 500.

## Fix

Catch the specific `Error` thrown by `mergeFrontmatterValue()` at the `frontmatter`
append/prepend call site and re-report it as `PatchFailed(ContentNotMergeable)`, which
maps to a `400`/`errorCode 40080` response. The guard is scoped narrowly:

- Only matches `error.constructor === Error` (the exact class thrown by
  `mergeFrontmatterValue`), so other failure types (e.g. a `YAMLError` from
  `yaml.stringify`) still surface as genuine 500s.
- Only applies to `targetType === "frontmatter"` with `append`/`prepend`.
- Preserves the original error via `.cause` for debugging.

## Testing

- Added `src/vaultOperations.test.ts` with 4 cases covering the type-mismatch scenario
  and confirming unrelated errors are unaffected.
- Full suite: 197/197 passing.
- Typecheck/lint clean.

### DIFF
--- a/src/vaultOperations.test.ts
+++ b/src/vaultOperations.test.ts
@@ -1,0 +1,82 @@
+import { VaultOperations, FileNotFoundError } from "./vaultOperations";
+import { App } from "../mocks/obsidian";
+import { PatchFailed, PatchFailureReason } from "markdown-patch";
+
+describe("VaultOperations.patchFileSection", () => {
+  let app: App;
+  let ops: VaultOperations;
+
+  const FIXTURE = ["---", "tags:", "  - theme/orig", "---", "", "Body text.", ""].join("\n");
+
+  beforeEach(() => {
+    app = new App();
+    app.vault._read = FIXTURE;
+    // @ts-ignore: mock App doesn't perfectly match the real Obsidian App interface
+    ops = new VaultOperations(app);
+  });
+
+  test("appending a type-mismatched scalar to a list frontmatter field throws PatchFailed(ContentNotMergeable), not a raw Error", async () => {
+    // Regression test for coddingtonbear/obsidian-local-rest-api#282: this used
+    // to escape markdown-patch's own applyPatch() as a bare
+    // `Error("Cannot merge objects of different types or unsupported types: object and string")`,
+    // which callers couldn't distinguish from a genuine server fault and which
+    // surfaced as an HTTP 500 / MCP error with no actionable message.
+    await expect(
+      ops.patchFileSection(
+        "somefile.md",
+        "frontmatter",
+        "tags",
+        "append",
+        "theme/added",
+        "application/json",
+      ),
+    ).rejects.toMatchObject({
+      constructor: PatchFailed,
+      reason: PatchFailureReason.ContentNotMergeable,
+    });
+  });
+
+  test("the PatchFailed thrown for a type mismatch preserves the original markdown-patch error as `cause`", async () => {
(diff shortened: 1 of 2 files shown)


coddingtonbear:
Thanks for the thorough write-up and the fix, @joshazmy! After digging into the root cause, the real fix belongs in `markdown-patch` itself: `mergeFrontmatterValue` should throw `MergeNotPossible` rather than a bare `Error`, which would let `applyPatch`'s existing catch handle the conversion to `PatchFailed(ContentNotMergeable)` automatically — the same path already used for other merge failures. I've filed that upstream at [coddingtonbear/markdown-patch#15](https://github.com/coddingtonbear/markdown-patch/issues/15).

Closing this in favor of the upstream fix. Once markdown-patch#15 is resolved and the dependency is bumped, the issue will resolve without needing a workaround here.

PR #310, opened 2026-07-23
Fix License section under Advanced settings not rendering

coddingtonbear:
## Summary

Same bug as the "Connection information" and "Certificate status" render-item bugs found and fixed on `feat/declarative-settings` (#307) before it merged — this one was in the same file but missed because it lives under "Advanced settings", which is hidden behind a toggle that's off by default, so it wasn't visible during that first pass of testing.

`setting.settingEl.remove()` followed by writing custom content into `group.listEl` detaches the framework-tracked element and puts new content somewhere the framework's own bookkeeping doesn't account for; some later pass (search indexing, most likely) re-inserts the original (still-intact, just detached) name/desc element in its place, and our content is nowhere to be found.

Fixed the same way as the other two: `setting.settingEl.empty()` instead of `.remove()`, and render into that same (still-attached) element with the existing `.full-width-setting` CSS class (`display: block`, overriding `.setting-item`'s default flex-row layout) rather than `group.listEl`.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm test` (242 passing)
- [x] `npm run build`
- [ ] Manual verification in Obsidian — no live Obsidian instance available in this environment. Please verify: toggle "Show advanced settings" on, and confirm the License section (paragraph + license link) renders correctly instead of a bare "License" label.


coddingtonbear:
@copilot resolve the merge conflicts in this pull request

PR #325, opened 2026-08-09
Keep the sample MCP config from overflowing its code block

rca-umb:
Simple edit to styles.css to fix text overflow.

The plain \<pre\> used for the sample MCP client JSON config had no
overflow handling, so long unbroken lines (specifically the
Authorization header line, which includes the full API key) spilled
past the block's background instead of scrolling or wrapping.

Every other pre in this settings page (the standalone copyable
Bearer-token/API-key values) already sets overflow-x: auto for this
same reason; this rule, which is the base style every plain pre in
the settings content falls back to, was the one place that got
missed.

I verified visually that this change fixes the issue.

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -30,6 +30,7 @@ div.obsidian-local-rest-api-content pre {
   background-color: var(--background-modifier-cover);
   font-family: monospace;
   user-select: all;
+  overflow-x: auto;
 }
 
 div.obsidian-local-rest-api-content p {


coddingtonbear:
Could I ask you to attach a "before" and "after" screenshot so I can understand what this fixes, concretely?

rca-umb:
Sure.  I'll also specify that I'm on MacOS 26.4.1 and Obsidian 1.13.4.

Here is what it looked like before:

Screenshot 2026-08-24 at 7 13 55 PM

And here is what it looks like after:

Screenshot 2026-08-24 at 7 16 08 PM

Note that in the after view, you can still scroll the code block to see the rest of the token, and if you click on the code block, it still selects everything.

This also happens in pre 5.1 versions as well, before the MCP config was moved to its own settings view.

coddingtonbear:
Awesome, thanks for the visual aid -- that helps a lot!

PR #326, opened 2026-08-16
Delete versions.json (#1)

zyntromedia:
The **Obsidian REST API schema** allows integration with other applications by providing endpoints for accessing and managing data within Obsidian. It supports operations like creating, reading, updating, and deleting notes, and is designed to work with standard data formats like JSON.


coddingtonbear:
:no_entry:

PR #335, opened 2026-08-26
Add vault_read_binary and vault_write_binary MCP tools

### LINKED ISSUE #284: Feature request: write binary files (attachments) through the MCP interface
### Summary

The MCP server bundled with this plugin exposes `vault_write` / `vault_patch`,
but their `content` is text only. There is currently no MCP tool to write a
**binary** file (PDF, image, etc.) into the vault. The only way today is to
bypass MCP and call the REST file-write endpoint directly with a binary body.
That works, but it's awkward for MCP-based agents.

### Current behavior

- `vault_write` takes a `content` string — fine for Markdown/text, but it
  cannot create binary attachments.
- To store e.g. a PDF I have to drop out of the MCP tool surface and hit the
  REST API directly:

```bash
curl -k -X PUT \
  'https://127.0.0.1:27124/vault/Attachments/report.pdf' \
  -H 'Content-Type: application/pdf' \
  --data-binary @report.pdf
```
This succeeds (HTTP 204).

Why this is awkward for MCP agents

An MCP client (LLM agent) normally only has the MCP tools available. To upload a
binary it must instead:

- know the REST base URL and port (and that HTTPS uses a self-signed cert, so
TLS verification has to be disabled),
- obtain and pass the API key out-of-band,
- build a raw HTTP request with a binary body,

all of which live outside the MCP abstraction the agent is otherwise usi
breaks the "everything through MCP tools" model and forces shell/HTTP access the
agent may not — and for safety often should not — have.

Proposed solution

Add binary support to the MCP interface, e.g. one of:

1. A new tool vault_write_binary (or upload_file) taking path,
content_base64, and optional content_type; it base64-decodes and writes
the bytes.
2. An optional encoding: "base64" (default "utf8") on the existing
vault_write, so the same tool can carry a binary payload.

Base64 keeps the MCP JSON transport happy while letting agents create
attachments without leaving the MCP tool surface.

Aside (possible docs mismatch)

Empirically, the working file path on my installed version is /vault/{path}.
Using /vault/files/{path} (as the published openapi.yaml seems to show)
created a literal files/ folder in the vault. Might just be a version skew —
flagging in case it's a docs bug.

Environment

- Plugin: Local REST API (with MCP) — v4.1.3
- Obsidian v1.12.7, macOS

coddingtonbear:
Closes #284.

## What this adds

Two MCP tools, `vault_read_binary` and `vault_write_binary`, so an agent that only has MCP tools can move attachments in and out of the vault at arbitrary paths.

- `vault_read_binary(path)` → `{ path, mimeType, size, encoding: "base64", content }`. `mimeType` is guessed from the extension the same way `GET /vault/<path>` guesses it, falling back to `application/octet-stream`.
- `vault_write_binary(path, content)` takes standard base64 and hands `writeFileContent` a `Buffer`, which already routes to `adapter.writeBinary`.

## Why, and what was actually broken

The REST API has handled binary content all along — `GET /vault/<path>` reads through `adapter.readBinary` with a mime-typed response (`src/requestHandler.ts:440`), and `PUT` accepts any content type because `express.raw({ type: "*/*" })` is the last body parser registered (`:2323`). None of that is mentioned in the README or `docs/src/`, which is probably why it keeps being asked for; this PR documents it.

MCP had no equivalent, and the failure mode was worse than the absence. `vault_read` runs `cachedRead` on whatever path it is given, so an attachment comes back as a lossy UTF-8 decode **with no error**, and `vault_list` enumerates attachments, so an agent is led right to them. Writing that string back destroys the file.

Issue #284 asked for exactly this, explicitly noting the REST `PUT` already worked, and was closed as completed five minutes after filing with no comment and no commit referencing it.

## Two deliberate constraints

**A 1 MiB ceiling on both tools** (`MaximumMcpBinaryBytes`). This is a context guard, not a storage limit: base64 in a tool argument or result passes through the model's context at roughly 0.35–0.45 tokens per byte, so a file a REST client would not think twice about is a five-figure token bill for an agent. The error names the REST endpoint, which is bounded only by `MaximumRequestSize` (1024mb). **The number is the thing in here most worth arguing with** — it is one constant.

**Strict base64 validation on write.** `Buffer.from(s, "base64")` discards anything it cannot decode instead of failing, so a mangled payload would land in the vault as silently-wrong bytes. The decoder checks length, alphabet, and then re-encodes and compares; base64url is rejected with a message that says so. A silently mangled attachment is worse than a failed call.

## Why this does not add risk

**The read path is additive.** `vault_read_binary` is a new tool; no existing tool's behavior changes. `readBinaryFileContent` is a new `VaultOperations` method that reuses `adapter.readBinary` — the same call `GET /vault/<path>` has always made — and adds the `TFile` existence check so its "file not found" answer matches `vault_read`'s.

**The write path reuses the REST write path exactly.** `writeFileContent(path, Buffer)` is the same call `PUT /vault/<path>` makes for a binary body, unchanged and already unit-tested (`src/vaultOperations.test.ts:75-79`). This PR adds no new way to write bytes to the vault; it adds a second caller of the existing one.

**Where it could plausibly be worse than before, and what bounds it:**

| Scenario | Bound |
|---|---|
| An agent pulls a huge attachment into context | The 1 MiB ceiling refuses before any bytes are read into a result, and points at REST |
| A malformed payload writes garbage over a real attachment | Strict validation rejects before `writeFileContent` is reached — asserted by four unit cases (bad length, bad alphabet, base64url, non-canonical) each also asserting `writeFileContent` was never called |
| Two more tools in every session's tool list | Real and unavoidable under this shape; the alternative considered was an `encoding` parameter on the existing tools |
| Base64 decode/encode mangles bytes | Round-tripped live against a real PNG, byte-identical (below) |

**Evidence, from the live plugin rather than from reasoning.** The rebuilt plugin was loaded in Obsidian and driven over its own MCP endpoint against the real vault:

- `vault_write_binary` on a 70-byte PNG (lead byte `0x89`, not valid UTF-8, so no path that decodes it as text could reproduce it) → `{"message":"OK","size":70}`
- `vault_read_binary` on the same path → `size: 70`, `mimeType: "image/png"`, and `content` **byte-identical** to what was written
- `vault_write_binary` with `-_-_` → rejected with the base64url message; nothing written
- Scratch file deleted afterward

**Honest gaps:**

- **The integration suite was not run.** The session that wrote this had no `OBSIDIAN_API_KEY`, so `npm run test:integration` could not start. The four integration tests added to `src/integration/mcp.test.ts` (round trip, REST/MCP agreement on the same bytes, invalid-base64 rejection, missing-file error) are **written but unexecuted** — they need a run before merge. The live probe above covers the same ground by hand for three of the four.
- **`vault_read` on a binary file is still silently lossy.** Guarding the text tools was evaluated and deliberately left out of this PR, since it changes existing tool behavior. It is worth its own change.
- **No `image`/`audio` content blocks.** A capable client would render those natively instead of receiving a wall of base64. Also deliberately out of scope here.
- The 1 MiB ceiling is not configurable. Adding a setting for it seemed like more surface than the problem warrants, but it is a constant, not a decision baked into the shape.

## Verification

- `npm test` — 433 passed, 9 suites. 10 new unit tests across `src/mcpHandler.test.ts` and `src/vaultOperations.test.ts`.
- `npm run typecheck`, `npx tsc -p tsconfig.integration.json --noEmit`, `npx eslint` on the touched files — all clean.
- `npm run build`, `npm run build-docs`, `npm run build-toc` — run and staged, per `AGENTS.md`.

## Layers touched

Per `AGENTS.md`, all in one commit: `src/mcpHandler.ts`, `src/vaultOperations.ts`, `src/constants.ts`, `README.md`, `docs/src/lib/descriptions/mcp.md`, `docs/src/lib/descriptions/vault-file-get.md`, `docs/src/openapi.jsonnet`, regenerated `docs/openapi.yaml`, unit tests, integration tests. `src/requestHandler.ts` is unchanged — REST already had the capability; what it lacked was documentation, which this adds.

One asymmetry is chosen rather than accidental: MCP gets a base64 convention that REST does not have, because REST carries raw bytes natively. Each protocol uses its natural carrier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ Give your scripts, browser extensions, and AI agents a direct line into your Obs
   * [Protocol revisions](#protocol-revisions)
   * [Connecting a client](#connecting-a-client)
   * [Available tools](#available-tools)
+  * [Binary files and attachments](#binary-files-and-attachments)
   * [Available resources](#available-resources)
 - [API Extensions](#api-extensions)
   * [Typed extension API](#typed-extension-api)
@@ -316,7 +317,9 @@ The exact config syntax varies by client; see the [Quick start](#mcp-clients) ex
 |---|---|
 | `vault_list` | List files and subdirectories inside a vault directory |
 | `vault_read` | Read a file's content, frontmatter, tags, and stat |
+| `vault_read_binary` | Read a file as raw bytes, base64-encoded, for attachments `vault_read` would corrupt |
 | `vault_write` | Create or overwrite a vault file |
+| `vault_write_binary` | Create or overwrite a vault file from base64-encoded raw bytes |
 | `vault_append` | Append content to the end of a vault file |
 | `vault_patch` | Patch a specific heading, block reference, or frontmatter field |
 | `vault_delete` | Delete a vault file (moves to trash by default) |
@@ -330,6 +333,14 @@ The exact config syntax varies by client; see the [Quick start](#mcp-clients) ex
 | `command_list` | List all registered Obsidian commands |
 | `command_execute` | Execute an Obsidian command by ID |
 | `open_file` | Open a file in the Obsidian UI |
+
+### Binary files and attachments
+
+The REST API has always handled binary content: `GET /vault/<path>` returns raw bytes with a `Content-Type` derived from the file extension, and `PUT /vault/<path>` accepts a body of any content type and stores it byte-for-byte. Neither has a practical size limit.
+
+MCP tools are a different story, because a tool's arguments and results pass through the model. `vault_read` and `vault_write` are text tools — they decode and encode UTF-8, which is lossy for anything that is not text — so reading an attachment with `vault_read` and writing the result back destroys the file. `vault_read_binary` and `vault_write_binary` exist for those files. They carry the bytes as standard base64 (not base64url), and `vault_write_binary` rejects a payload that is not the canonical encoding of the bytes it decodes to rather than writing mangled ones.
+
+Because base64 costs roughly 0.35-0.45 tokens per byte of context, both binary tools refuse files over 1 MiB. That ceiling is a context guard, not a storage limit — move larger attachments over the REST endpoints above.
 
 ### Available resources
 
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -1440,7 +1440,9 @@ paths:
         |---|---|
         | `vault_list` | List files and subdirectories inside a vault directory |
         | `vault_read` | Read a file's full content, frontmatter, tags, and stat |
+        | `vault_read_binary` | Read a file as raw bytes, base64-encoded, for attachments `vault_read` would corrupt |
         | `vault_write` | Create or overwrite a vault file |
+        | `vault_write_binary` | Create or overwrite a vault file from base64-encoded raw bytes |
         | `vault_append` | Append content to the end of a vault file |
         | `vault_patch` | Patch a specific heading, block reference, or frontmatter field |
         | `vault_delete` | Delete a vault file (moves to trash by default) |
@@ -1454,6 +1456,12 @@ paths:
         | `command_list` | List all registered Obsidian commands |
         | `command_execute` | Execute an Obsidian command by ID |
         | `open_file` | Open a file in the Obsidian UI |
+        
+        ### Binary files
+        
+        `vault_read` and `vault_write` are text tools: they decode and encode UTF-8, which is lossy for anything that is not text, so reading an attachment through `vault_read` and writing the result back destroys the file. Use `vault_read_binary` and `vault_write_binary` for attachments instead. Both carry the file as standard base64 (not base64url), and `vault_write_binary` rejects a payload that is not the canonical encoding of the bytes it decodes to rather than writing mangled ones.
+        
+        Base64 in a tool argument or result passes through the model's context at roughly 0.35-0.45 tokens per byte, so both tools refuse files over 1 MiB. That is a context guard rather than a storage limit: `GET` and `PUT /vault/{filename}` carry raw bytes of any size at no context cost, and are the right way to move a large attachment.
         
         ## Available resources
         
@@ -2066,6 +2074,8 @@ paths:
     get:
       description: |
         Returns the content of the file at the specified path in your vault should the file exist.
+        
+        Binary files are served as raw bytes, with a `Content-Type` derived from the file extension, so this endpoint reads attachments — images, PDFs, audio — as well as notes. There is no size limit on the response. MCP clients have `vault_read_binary` for the same job, but it base64-encodes the file into the model's context and so caps the file size; anything larger belongs here.
         
         # Targeting a Sub-part of your Document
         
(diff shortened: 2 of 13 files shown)


coddingtonbear:
## Integration evidence

**The integration suite did not run.** `npm run test:integration` aborted before any test executed — all 12 suites failed at import with `OBSIDIAN_API_KEY env var is required to run integration tests.` (`src/integration/client.ts:5`). The headless session that produced this branch has no API key in its environment. So the four tests added to `src/integration/mcp.test.ts` are **written but unexecuted**, and someone should run them before this merges.

What was done instead: the branch was built (`npm run build`), Obsidian picked the rebuilt plugin up, and the two new tools were driven over the plugin's own live MCP endpoint against the real vault. That is a weaker guarantee than a suite run — it is four calls, made by hand, not a repeatable assertion — but it exercises the same code path end to end, including Obsidian's real vault adapter.

### What was observed

Fixture: a 70-byte 1×1 PNG. Chosen because its lead byte is `0x89`, which is not valid UTF-8 — no path that decodes the file as text could reproduce it, so a successful round trip proves the bytes never went through a string.

| Call | Result |
|---|---|
| `vault_write_binary` → `attachments/claude-binary-tool-probe.png` | `{"message":"OK","size":70}` |
| `vault_read_binary` on the same path | `size: 70`, `mimeType: "image/png"`, `encoding: "base64"`, `content` **byte-identical** to the payload written |
| `vault_write_binary` with `content: "-_-_"` | Rejected: `content must be standard base64-encoded bytes (contains characters outside the base64 alphabet)...`. Nothing written. |
| `vault_delete` on the scratch path | `{"message":"OK"}` — no residue left in the vault |

That covers three of the four new integration tests by hand: the round trip, the invalid-base64 rejection, and (via the write) file creation. **Not covered by the probe:** the REST/MCP agreement test, which asserts `GET /vault/<path>` returns the same bytes MCP wrote. That one is only asserted by the unexecuted integration test, and by the fact that both layers call the same `adapter.readBinary`/`writeBinary` pair.

### Unit suite

`npm test` — **433 passed, 9 suites, 0 failed.** 10 of those are new:

- `vault_read_binary`: base64 payload with mime type and size; `application/octet-stream` fallback for an unknown extension; refusal over the ceiling.
- `vault_write_binary`: Buffer handed to `writeFileContent` with bytes intact; whitespace-wrapped base64 accepted; four rejection cases (bad length, character outside the alphabet, base64url, non-canonical encoding), each also asserting `writeFileContent` was never called; refusal over the ceiling; empty payload writes an empty file.
- `readBinaryFileContent`: returns the adapter's bytes rather than a decoded string; throws `File not found` for a non-file path, matching `vault_read`.

Also clean: `npm run typecheck`, `npx tsc -p tsconfig.integration.json --noEmit`, and `npx eslint` over every touched file.

### New risk, and what bounds it

- **Oversized reads into model context** — the 1 MiB ceiling refuses before bytes reach a result. Asserted by unit test in both directions. The ceiling value itself is unvalidated against real-world use; it is one constant and easy to change.
- **Mangled bytes written over a real attachment** — strict validation rejects before `writeFileContent` is reached, asserted four ways. The re-encode-and-compare check is stricter than most decoders; if a client emits non-canonical base64 it will get an error rather than a corrupt file, which is the intended trade.
- **Two more entries in every session's tool list** — real, unavoidable under this shape, and the reason the alternative (an `encoding` parameter on the existing tools) was on the table.

### Unchanged by design

`vault_read` on a binary path still returns a lossy UTF-8 decode rather than an error. This branch adds a safe path; it does not close the unsafe one. That is a behavior change to existing tools and belongs in its own PR.

PR #336, opened 2026-08-26
Expose response headers to browser clients via CORS

coddingtonbear:
Both `cors()` calls were configured with `methods` only. The `cors` package omits `Access-Control-Expose-Headers` entirely when `exposedHeaders` is unset and never defaults it, so a browser client may read only the seven CORS-safelisted response headers. Every header this API sets to tell a client something is on the wire and readable by curl, but invisible to JavaScript.

## What that costs today

| Header | Sites | What a browser client loses |
|---|---|---|
| `Mcp-Session-Id` | MCP SDK transport | **Sessionful MCP cannot hand-shake** — the SDK's own client reads `response.headers.get('mcp-session-id')` and gets `null` |
| `Content-Location` | 9 | "Where did my write actually land" — the point of the `/active/` and `/periodic/` routes |
| `Markdown-Patch-Warnings` | 1 | A PATCH's warnings vanish; the call looks perfectly clean |
| `ETag` | Express, automatic | No conditional requests, no optimistic concurrency |
| `Content-Disposition` | 1 | Cannot honour the suggested filename |
| `Deprecation` | 4 | The markdown-patch 1.x sunset warning does not reach them |
| `X-Response-Time` | 1 | Unreadable |

`Mcp-Session-Id` is what moves this from a missed courtesy to a broken feature. The rest were degraded quietly.

This project explicitly has browser consumers — the web extension — so this is not a hypothetical class of client.

## The change

One shared `CorsOptions` constant, carrying `exposedHeaders: "*"`, used by both the main API router and the MCP router.

The wildcard was chosen over an enumerated list because nothing would enforce an explicit list staying in sync, and two of the affected headers (`ETag`, `Mcp-Session-Id`) are set by Express and the MCP SDK rather than by this codebase — they would sit in such a list as literals regardless.

## Why this does not add risk

The change adds a response header to responses that previously had none of it. It cannot alter status codes, bodies, or the set of headers actually sent — only whether a browser is permitted to read headers that were already there. Concretely, the scenarios where it could plausibly be *worse* than the status quo, and what bounds each:

- **Credentialed requests.** Per the Fetch standard the wildcard is only special for requests made *without* credentials; a credentialed request reads it as the literal header name `*`. Bounded because this API sends `Access-Control-Allow-Origin: *` and never `Access-Control-Allow-Credentials`, so a credentialed request already cannot succeed here. Recorded in the constant's doc comment: if the origin policy ever gains credentials, this must become an explicit list at the same time.
- **Old Safari.** The wildcard is honoured only from Safari 15.4 (March 2022). Bounded because older clients fall back to the CORS safelist — exactly where they were before this change. No regression, just no fix.
- **Information exposure.** The wildcard exposes response headers to script on a cross-origin page, but `Set-Cookie` is excluded by the Fetch standard and this API sets no other header carrying a secret; the API key travels in the request direction, never the response. Any origin could already read the full response *body* — `Access-Control-Allow-Origin: *` with bearer auth means an attacker holding the key needs no browser at all, and one without it gets a 401 whose headers say nothing.
- **The request direction.** Untouched, and pinned by a test in both the unit and integration suites, so a later change to the response direction cannot quietly cost us header reflection on preflight.

## Verification

TDD per AGENTS.md: the failing tests are their own commit (`a63c178`), the fix follows.

- **Unit — run, passing.** 3 new tests in `src/requestHandler.test.ts`. Full suite 423/423 green, 420 before. The two response-direction tests were confirmed to fail against `main` and pass after the fix; the preflight test passed both before and after, which is what establishes that the request direction was never the defect.
- **`npm run typecheck`, `npm run lint`, `npm run build`** all clean. The 4 lint warnings are pre-existing sentence-case warnings in the settings UI, untouched here.
- **Integration — written, NOT run.** 4 new tests in `src/integration/cors.test.ts`, including one that performs a real `POST` returning `Content-Location` and asserts the header and the directive together. See the gap below.

## Honest gaps

- **The integration suite never ran.** No `OBSIDIAN_API_KEY` in the session that wrote this, so all 13 suites fail at import. The 4 new tests are unexecuted and **want a run before this merges.**
- **No live browser check.** The `Mcp-Session-Id` consequence is read out of the MCP SDK's source, not watched in a browser. The claim that this fixes sessionful MCP for browser clients is therefore reasoned, not observed — it is the one assertion here I would most want confirmed by someone with a browser.
- **`docs/src/` is untouched.** AGENTS.md asks that REST, MCP, OpenAPI and README move together, but exposed headers are a transport-level concern belonging to no single endpoint and there is no obvious OpenAPI home for them. Documented in a new README section instead, with the TOC regenerated. Say the word if you would rather it were documented per-endpoint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ Give your scripts, browser extensions, and AI agents a direct line into your Obs
     + [Cursor](#cursor)
     + [Other clients](#other-clients)
 - [API overview](#api-overview)
+  * [Browser clients and response headers](#browser-clients-and-response-headers)
 - [Patching notes](#patching-notes)
   * [Raw-content mode](#raw-content-mode)
 - [Targeting specific sections](#targeting-specific-sections)
@@ -171,6 +172,12 @@ Any MCP client that supports the Streamable HTTP transport can connect to `https
 | `/mcp/` | GET POST | MCP (Model Context Protocol) server — connect AI agents directly to your vault |
 
 For full request/response details, see the [interactive docs](https://coddingtonbear.github.io/obsidian-local-rest-api/).
+
+### Browser clients and response headers
+
+Several endpoints answer in a response header rather than in the body: `Content-Location` tells you where a write actually landed, `Markdown-Patch-Warnings` reports what a `PATCH` had to work around, `Deprecation` warns that a format is sunsetting, and `Mcp-Session-Id` carries the session for a sessionful MCP connection.
+
+Browsers hide response headers from JavaScript unless the server opts them in, so the API sends `Access-Control-Expose-Headers: *` and all of them are readable with `response.headers.get(...)`. Safari honours the wildcard from 15.4 onward; older browsers see only the [CORS-safelisted headers](https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_response_header). Requests made with `credentials: "include"` are not supported — the API authenticates with a bearer token and sends `Access-Control-Allow-Origin: *`, which browsers reject for credentialed requests.
 
 ## Patching notes
 
--- a/src/integration/cors.test.ts
+++ b/src/integration/cors.test.ts
@@ -1,0 +1,71 @@
+import { authedFetch, unauthFetch, ensureServerReachable, resetFixture, deleteFixture, BASE_URL } from "./client";
+import { FIXTURE_DOCUMENT, TEST_PATH, HEADING_DELTA } from "./fixtures";
+
+// The unit tests assemble the router under supertest; these drive the real server, which
+// is where the ordering of cors() against the auth and body-parsing middleware actually
+// holds. A browser reads a response header only if Access-Control-Expose-Headers names
+// it, and this API answers several questions in headers rather than in the body.
+
+const ORIGIN = "https://example.com";
+
+beforeAll(async () => {
+  await ensureServerReachable();
+});
+
+describe("CORS exposed response headers", () => {
+  test("GET / exposes every response header", async () => {
+    const res = await unauthFetch("/", { headers: { Origin: ORIGIN } });
+    expect(res.status).toBe(200);
+    expect(res.headers.get("access-control-expose-headers")).toBe("*");
+  });
+
+  test("the /mcp/ router exposes every response header", async () => {
+    // cors runs ahead of the MCP router's auth middleware, so an unauthenticated request
+    // is enough to observe the directive -- and it avoids needing a valid MCP handshake.
+    // Mcp-Session-Id rides this router, and the MCP SDK's own browser client reads that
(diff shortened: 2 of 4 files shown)


coddingtonbear:
## Test evidence

### The integration suite did not run

`npm run test:integration` fails before executing a single test:

```
● Test suite failed to run
    OBSIDIAN_API_KEY env var is required to run integration tests.
      at Object.<anonymous> (src/integration/client.ts:5:9)

Test Suites: 13 failed, 13 total
Tests:       0 total
```

This was authored in a headless session with no API key and no live Obsidian, so **all 13 suites** fail at import — not just the new one. The 4 tests in `src/integration/cors.test.ts` are written but unexecuted. They want a run before merge.

### Unit suite — run, passing

```
Test Suites: 9 passed, 9 total
Tests:       423 passed, 423 total
```

420 before this branch; 3 added.

### The regression tests were proven to catch the defect

This is the part that makes the unit coverage worth something. The tests were written and committed *before* the fix (`a63c178`), and run against unmodified `main`:

```
● requestHandler › cors › GET / exposes every response header to browser clients
    Expected: "*"
    Received: undefined

● requestHandler › cors › the /mcp/ router exposes every response header to browser clients
    Expected: "*"
    Received: undefined

Tests: 2 failed, 285 skipped, 1 passed, 288 total
```

Two facts fall out of that run:

1. Both response-direction tests fail on `main` and pass after `72c92a3`. They are pinned to the actual defect, not to incidental behaviour.
2. The third test — the preflight one — **passed before the fix as well.** That is the evidence for the claim in the description that this is a response-direction-only defect: `cors` was already reflecting `Access-Control-Request-Headers` correctly, and nothing here changes that.

### Other checks

| Check | Result |
|---|---|
| `npm run typecheck` | clean |
| `npm run lint` | clean — 4 warnings, all pre-existing sentence-case warnings in the settings UI |
| `npm run build` | clean, no generated-file drift (`publicApi.d.ts`/`publicApi.js` unchanged) |
| `npm run build-toc` | regenerated; one new README entry |

### What this evidence does not cover

- **No browser was involved.** Every assertion here is made by `supertest` or by reading source. The headline claim — that a browser MCP client can now complete a sessionful handshake — is inferred from the MCP SDK reading `response.headers.get('mcp-session-id')`, not observed. A five-minute check in a real browser would settle it.
- **`exposedHeaders: "*"` is asserted as a literal string.** The tests confirm the directive is sent; they cannot confirm any particular browser honours the wildcard, which is why the Safari 15.4 floor is documented rather than tested.
- **A live check against the running plugin was attempted and abandoned.** `curl` against `127.0.0.1:27124` needed an approval the headless session could not obtain, so no on-the-wire capture backs this up the way one backs the original diagnosis.

Issue #327, opened 2026-08-17
NoteJson links/backlinks/unresolvedLinks can be served before vault-wide link resolution completes


didierhk:
First, thank you for `unresolvedLinks` (#302 / #305). It let us drop an entire subprocess backend from our client.

## Summary

`getFileMetadataObject` serves `links`, `backlinks` and `unresolvedLinks` as soon as the *per-file* metadata cache for the requested note is available. All three are derived from vault-global relations, so they are not final until every file in the vault has been indexed. A client cannot distinguish a partial answer from a complete one.

## Where (tag 5.1.0, `src/vaultOperations.ts`)

- `getFileMetadataObject` awaits `waitForFileCache(file)` at line 215
- reads `this.app.metadataCache.resolvedLinks[file.path] ?? {}` at 232 and `this.app.metadataCache.unresolvedLinks[file.path] ?? {}` at 235
- includes both unconditionally in the returned object at 249
- `buildBacklinksIndex` (198) walks `this.app.metadataCache.resolvedLinks` across the whole vault

`waitForFileCache` (81-124) resolves on `metadataCache.on("changed")` for that one file, or times out after 5000 ms and returns whatever `getFileCache` yields at that moment.

## Why the existing wait does not cover this

The wait is per-file; the data is vault-global. Whether `[[Foo]]` lands in `links` or in `unresolvedLinks` depends on whether `Foo.md` has been indexed, not on whether the *requested* note has. During startup, a note whose own cache is ready can therefore report a link as unresolved purely because its target has not been parsed yet. `buildBacklinksIndex` has the same property, since a backlink is missing until the linking file is indexed.

Obsidian exposes `metadataCache.on("resolved")` for this case, documented as firing when all files have been resolved.

## Why it matters to a consumer

We use the plugin as the data source for an MCP server that gives LLM agents a "find broken links" tool. An agent that asks and gets a short list has no way to tell a clean vault from a warming cache. Absence of results is indistinguishable from absence of the problem, so we mark every graph result unverified and document that it cannot support a "no broken links" conclusion, a compliance sign-off, or automated rename and delete repair. That is the honest labelling, but it means the data cannot be used for the decisions people most want it for.

## Request

Either of these would resolve it, and the second is the cheaper one:

1. Gate these three fields on vault-wide resolution having completed at least once since load.
[...]

PR #337, opened 2026-08-26
Cache the vault backlinks index instead of rebuilding it per read

coddingtonbear:
Every single-note metadata read walked the entire vault link graph to extract the backlinks of one file. This builds that index at most once per change to the graph.

## The defect

`getFileMetadataObject` takes an optional prebuilt index and falls back to building one when a caller does not pass it (`src/vaultOperations.ts`). The bulk search path passes one and threads it through its loop; the single-note paths never do:

- `GET /vault/<path>` with `Accept: application/vnd.olrapi.note+json` (`src/requestHandler.ts:458-461`)
- the MCP `vault_read` tool (`src/mcpHandler.ts:490`)

So each of those walked all of `metadataCache.resolvedLinks` and every target within it, kept one entry, and discarded the rest.

## The change

`getBacklinksIndex()` memoises the index; the memo is dropped whenever Obsidian says the graph moved. `buildBacklinksIndex()` itself is untouched, and the optional argument stays — a bulk caller wants one snapshot across its whole loop so that every row of a result set describes the same moment even if the graph moves mid-loop.

## Why this does not add risk

The saving is not the interesting part; **staleness** is. A cached graph that outlives a change to the real one serves backlinks that are quietly wrong — a worse failure than the cost it removes. Three things bound that:

**1. Invalidation is deliberately broad.** `metadataCache`'s `resolve` (per file), `resolved` (per batch), and `deleted`, plus the vault's `rename` and `delete`. `resolvedLinks` is keyed by path, and `metadataCache` documents that it does not announce renames ("This is not called when a file is renamed for performance reasons. You must hook the vault rename event for those"), so the vault events are not redundancy for its own sake. An invalidation too many costs a scan that used to be paid on *every* request; one too few is a correctness bug.

**2. Each listener is pinned by a test that fails without it.** Verified by removing them one at a time — see the evidence comment.

**3. Observed against a real vault, not just asserted.** The build was run against a live Obsidian and the create-link / rename / delete cycle driven through the plugin's own MCP tools. Numbers and transcript in the evidence comment.

Two smaller defensive calls:

- **The backlinks array handed out is now a copy.** The cached index outlives the response built from it, so returning its own array would let one caller's mutation reach every caller after it. Previously each response owned a freshly built array, so this hazard is created by the cache and closed in the same change.
- **`VaultOperations.dispose()`** releases the listeners, called from `onunload`. It matters only at unload, but a listener left behind holds the instance alive.

## Where it could still be worse than before

- **A graph mutation Obsidian never announces** would now persist instead of being picked up by the next request's fresh scan. Every mutation path I could find announces itself, and all five announcements are handled — but this is the failure mode to look for, and it is why invalidation is broad rather than clever.
- **The window between `resolvedLinks` being mutated and the event firing** is served from the memo, where it used to be served from a fresh read of the live object. Obsidian mutates and then fires, so the window is synchronous, and the uncached path was equally "as of the last resolution" — but it is not literally identical behaviour.
- **Memory**: one index for the vault, the same object the search path already built per request.

## Verification

429 unit tests pass (10 new), `typecheck`, `lint` and `build` clean. Live behaviour driven against a real vault. **The integration suite did not run** — no `OBSIDIAN_API_KEY` in this session; details in the evidence comment.

### DIFF
--- a/mocks/obsidian.ts
+++ b/mocks/obsidian.ts
@@ -117,6 +117,34 @@ export class Vault {
     file.basename = (path.split("/").pop() ?? path).replace(/\.md$/, "");
     return file;
   }
+
+  _listeners: Map<string, ((...data: unknown[]) => unknown)[]> = new Map();
+
+  on(event: string, callback: (...data: unknown[]) => unknown): void {
+    if (!this._listeners.has(event)) {
+      this._listeners.set(event, []);
+    }
+    this._listeners.get(event)!.push(callback);
+  }
+
+  off(event: string, callback: (...data: unknown[]) => unknown): void {
+    const listeners = this._listeners.get(event);
+    if (listeners) {
+      const index = listeners.indexOf(callback);
+      if (index !== -1) {
+        listeners.splice(index, 1);
+      }
+    }
+  }
+
+  // Helper method for tests to simulate vault events -- `rename`, `delete` --
+  // with whatever arguments Obsidian would pass.
+  _emit(event: string, ...data: unknown[]): void {
+    const listeners = this._listeners.get(event);
+    if (listeners) {
+      listeners.forEach((cb) => cb(...data));
+    }
+  }
 }
 
 class FileManager {
@@ -198,9 +226,15 @@ export class MetadataCache {
   // file's new content as the second argument, which callers use to tell an update
   // for their own write apart from one for an unrelated revision.
   _emitChanged(file: TFile, data = ""): void {
-    const listeners = this._listeners.get("changed");
+    this._emit("changed", file, data);
+  }
+
+  // Fires any other cache event by name -- `resolve`, `resolved`, `deleted` --
+  // with whatever arguments Obsidian would pass.
+  _emit(event: string, ...data: unknown[]): void {
+    const listeners = this._listeners.get(event);
     if (listeners) {
-      listeners.forEach((cb) => cb(file, data));
+      listeners.forEach((cb) => cb(...data));
     }
   }
 }
--- a/src/main.ts
+++ b/src/main.ts
@@ -249,6 +249,7 @@ export default class LocalRestApi extends Plugin {
 
   onunload() {
     this.requestHandler?.mcpHandler.close();
+    this.requestHandler?.operations.dispose();
     if (this.secureServer) {
       this.secureServer.closeAllConnections();
       this.secureServer.close();
--- a/src/vaultOperations.test.ts
+++ b/src/vaultOperations.test.ts
@@ -146,3 +146,135 @@ describe("simpleSearch surrogate handling", () => {
     expect(await results()).toEqual([{ context: "xneedle🔌" }]);
   });
 });
+
+// ---------------------------------------------------------------------------
+// A single-note read must not rescan the whole vault link graph.
+//
+// buildBacklinksIndex walks every entry of metadataCache.resolvedLinks and every
+// target within it -- the entire vault link graph -- to answer "which files link
+// here". The bulk search path builds it once and threads it through its loop, but
+// every single-note metadata read (GET /vault/<path> as note+json, and the MCP
+// vault_read tool) used to pay for a fresh whole-graph scan and then discard all
+// but one entry of the result.
+//
+// Caching it makes correctness the interesting question rather than cost: a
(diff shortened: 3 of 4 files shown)


coddingtonbear:
## Test evidence

### What ran

| | |
|---|---|
| `npm test` | **429 passed**, 9 suites, 0 failed (10 new tests) |
| `npm run typecheck` | clean |
| `npm run lint` | 0 errors, 4 warnings — all pre-existing `sentence-case` warnings in `src/main.ts` UI strings, none on changed lines |
| `npm run build` | clean, including `build-types` |
| `npm run test:integration` | **did not run** — see gaps |

### The tests are load-bearing, not decorative

Each invalidation listener was removed one at a time and the suite re-run. Removing `metadataCache.on("resolve", ...)`:

```
✓ repeated single-note reads scan the link graph once
✕ metadataCache resolve invalidates the cached index
✓ metadataCache resolved invalidates the cached index
✓ metadataCache deleted invalidates the cached index
✓ vault rename invalidates the cached index
✓ vault delete invalidates the cached index
Tests: 1 failed, 16 passed
```

Exactly one test failed, and it was the matching one. Same for the array copy: dropping the spread fails `a caller that mutates the backlinks it was handed cannot corrupt the cache` and nothing else.

The five invalidation tests were also committed *before* the cache existed (90eb404), where they pass trivially — they exist to fail the moment someone adds a cache without one of them.

### Measured cost, synthetic harness

A throwaway jest harness built synthetic `resolvedLinks` graphs and timed the real `VaultOperations` methods (20 timed builds after warmup; 50 single-note `getFileMetadataObject` calls per arm, invalidating between reads in the "before" arm to reproduce the old behaviour). Node 20 on Linux, not Electron.

| Vault | `buildBacklinksIndex()` median | 50 reads, rebuilding each time | 50 reads, cached |
|---|---|---|---|
| 1,000 notes / 5,000 link entries | 0.43 ms | 21.3 ms (0.43 ms/read) | 0.5 ms (0.011 ms/read) |
| 9,000 notes / 45,000 link entries | 4.97 ms | 267.7 ms (5.35 ms/read) | 6.8 ms (0.136 ms/read) |
| 9,000 notes / 180,000 link entries | 16.16 ms | 808.2 ms (16.16 ms/read) | 16.2 ms (0.324 ms/read) |

9,000 notes is the vault size reported in #327. The scan scales with link entries, so the middle row is the conservative reading and the last row is a densely linked vault of that size.

The harness is not in the PR: it measures a mock vault rather than Obsidian's, so committing it would pin a number that means little on other hardware. The shape — linear in link entries, paid once per graph change rather than once per request — is the claim.

### Observed against a real vault

The branch was built into a live Obsidian and driven through the plugin's own MCP `vault_read`, which reaches `getFileMetadataObject` by the same path as `GET /vault/<path>` as `note+json`. Each read after the first would serve a stale answer if invalidation were missing:

1. created `claude-scratch-backlinks-target.md`, read it → `backlinks: []` *(cache now warm, and warm with the wrong answer for what comes next)*
2. created `claude-scratch-backlinks-source.md` linking to it, read the target → `backlinks: ["claude-scratch-backlinks-source.md"]`
3. renamed the source to `claude-scratch-backlinks-source-renamed.md`, read the target → `backlinks: ["claude-scratch-backlinks-source-renamed.md"]`
4. deleted the source, read the target → `backlinks: []`

Step 3 is the one worth noting: it is the case the unit tests could only assume, since `metadataCache` deliberately stays quiet about renames. Both scratch files were deleted afterward.

### Honest gaps

- **The integration suite never ran.** No `OBSIDIAN_API_KEY` in this session, so all suites fail at import. This change adds no endpoint behaviour and no integration tests, but the existing `note+json` and `vault_read` coverage is exactly what would catch a regression here and it wants a run before merge.
- **The timings are synthetic.** Mock vault, Node rather than Electron, and `getFileMetadataObject`'s other work (`cachedRead`, frontmatter) is mocked away — so the per-read totals are the graph-walking component, not a whole request. The ratio is the point.
- **No real large vault was measured.** The live checks above prove correctness, not speed; the vault they ran against is not 9,000 notes.
- **"Every mutation announces itself" is read out of the API docs and confirmed for four paths live.** A sixth path that silently rewrites `resolvedLinks` would defeat the cache, and I cannot prove one does not exist.

PR #343, opened 2026-08-28
Cache the vault backlinks index instead of rebuilding it per read

coddingtonbot:
Every single-note metadata read walked the entire vault link graph to extract the backlinks of one file. This builds that index at most once per change to the graph.

## The defect

`getFileMetadataObject` takes an optional prebuilt index and falls back to building one when a caller does not pass it (`src/vaultOperations.ts`). The bulk search path passes one and threads it through its loop; the single-note paths never do:

- `GET /vault/<path>` with `Accept: application/vnd.olrapi.note+json` (`src/requestHandler.ts:458-461`)
- the MCP `vault_read` tool (`src/mcpHandler.ts:490`)

So each of those walked all of `metadataCache.resolvedLinks` and every target within it, kept one entry, and discarded the rest.

## The change

`getBacklinksIndex()` memoises the index; the memo is dropped whenever Obsidian says the graph moved. `buildBacklinksIndex()` itself is untouched, and the optional argument stays — a bulk caller wants one snapshot across its whole loop so that every row of a result set describes the same moment even if the graph moves mid-loop.

## Why this does not add risk

The saving is not the interesting part; **staleness** is. A cached graph that outlives a change to the real one serves backlinks that are quietly wrong — a worse failure than the cost it removes. Three things bound that:

**1. Invalidation is deliberately broad.** `metadataCache`'s `resolve` (per file), `resolved` (per batch), and `deleted`, plus the vault's `rename` and `delete`. `resolvedLinks` is keyed by path, and `metadataCache` documents that it does not announce renames ("This is not called when a file is renamed for performance reasons. You must hook the vault rename event for those"), so the vault events are not redundancy for its own sake. An invalidation too many costs a scan that used to be paid on *every* request; one too few is a correctness bug.

**2. Each listener is pinned by a test that fails without it.** Verified by removing them one at a time — see the evidence comment.

**3. Observed against a real vault, not just asserted.** The build was run against a live Obsidian and the create-link / rename / delete cycle driven through the plugin's own MCP tools. Numbers and transcript in the evidence comment.

Two smaller defensive calls:

- **The backlinks array handed out is now a copy.** The cached index outlives the response built from it, so returning its own array would let one caller's mutation reach every caller after it. Previously each response owned a freshly built array, so this hazard is created by the cache and closed in the same change.
- **`VaultOperations.dispose()`** releases the listeners, called from `onunload`. It matters only at unload, but a listener left behind holds the instance alive.

## Where it could still be worse than before

- **A graph mutation Obsidian never announces** would now persist instead of being picked up by the next request's fresh scan. Every mutation path I could find announces itself, and all five announcements are handled — but this is the failure mode to look for, and it is why invalidation is broad rather than clever.
- **The window between `resolvedLinks` being mutated and the event firing** is served from the memo, where it used to be served from a fresh read of the live object. Obsidian mutates and then fires, so the window is synchronous, and the uncached path was equally "as of the last resolution" — but it is not literally identical behaviour.
- **Memory**: one index for the vault, the same object the search path already built per request.

## Verification

429 unit tests pass (10 new), `typecheck`, `lint` and `build` clean. Live behaviour driven against a real vault. **The integration suite did not run** — no `OBSIDIAN_API_KEY` in this session; details in the evidence comment.

### DIFF
--- a/src/vaultOperations.ts
+++ b/src/vaultOperations.ts
@@ -8,6 +8,7 @@ import {
   prepareSimpleSearch,
   TFile,
 } from "obsidian";
+import type { Events } from "obsidian";
 import path from "path";
 import {
   applyPatch,
@@ -47,45 +48,24 @@ import {
 import { toArrayBuffer } from "./utils";
 
 /**
- * Every event Obsidian's metadata cache publicly declares, and every event its
- * vault publicly declares.
- *
- * Deliberately the whole surface rather than the subset that looks like it moves
- * the link graph. `resolvedLinks` is Obsidian's own derived state, and which of
- * its events happen to precede an update to it is an implementation detail of a
- * dependency -- betting a cache's correctness on having read that detail right
- * is a bet nobody can win permanently. Subscribing to everything removes the
- * judgement call, and costs only a rebuild the uncached code performed on every
- * single read anyway.
- *
- * `src/vaultOperations.test.ts` reads these names back out of the installed
- * obsidian typings and fails if the two ever disagree, so an Obsidian upgrade
- * that adds an event is a red test rather than a cache that quietly goes stale.
- */
-export const METADATA_CACHE_EVENTS = [
-  "changed",
-  "deleted",
-  "resolve",
-  "resolved",
-] as const;
-export const VAULT_EVENTS = ["create", "modify", "delete", "rename"] as const;
-
-/**
  * How long a built backlinks index may be served before it is rebuilt anyway.
  *
- * The listeners above cover everything Obsidian announces, but "everything it
- * announces" is not the same as "everything that happens": an event added in a
- * future release, or an internal path that rewrites `resolvedLinks` without
- * saying so, would otherwise leave a stale index in place for as long as the
- * plugin runs. Ageing the index out turns that unbounded failure into a bounded
- * one, without depending on any part of Obsidian's API being what we think.
+ * Trigger interception (see `interceptTriggers`) covers everything Obsidian
+ * announces, whatever the announcement is called. What it cannot cover is a
+ * change that announces nothing at all -- an internal path that rewrites
+ * `resolvedLinks` in silence -- which would otherwise leave a stale index in
+ * place for as long as the plugin runs. Ageing the index out turns that
+ * unbounded failure into a bounded one without depending on any part of
+ * Obsidian's API being what we think it is.
  *
- * A second is chosen so the ceiling costs nothing that matters: the behaviour
- * this cache replaced rebuilt the index on *every* read, so even the worst case
- * here -- one rebuild per second -- stays far below what the endpoint used to
- * pay, while the burst of reads the cache exists for still scans once.
+ * The ceiling is the backstop, not the mechanism, so it is set where it costs
+ * least: a rebuild scans the vault's whole link graph, measured at ~5 ms for a
+ * 9,000-note vault and ~16 ms at four times that, and it runs on the same
+ * thread as Obsidian's UI. A minute keeps that off the critical path of a
+ * sustained read load entirely, while still bounding a silent change at a
+ * minute rather than a session.
  */
-export const BACKLINKS_INDEX_MAX_AGE_MS = 1000;
+export const BACKLINKS_INDEX_MAX_AGE_MS = 60_000;
 
 /**
  * Writes go through Vault.modify/Vault.create rather than Vault.adapter.write.
@@ -101,6 +81,8 @@ export class VaultOperations {
   private cachedBacklinksIndex: Record<string, string[]> | null = null;
   private cachedBacklinksIndexBuiltAt = 0;
 
+  private readonly restoreTriggers: (() => void)[] = [];
+
   /**
    * Dropped whenever Obsidian says anything at all has happened.
    *
@@ -114,18 +96,8 @@ export class VaultOperations {
   };
 
   constructor(readonly app: App, readonly settings: LocalRestApiSettings) {
-    // Obsidian types `on` as one overload per literal event name, so iterating a
-    // union of them needs a cast to any single literal to typecheck. The handler
-    // takes no arguments, which every one of those callback signatures accepts.
-    for (const event of METADATA_CACHE_EVENTS) {
-      this.app.metadataCache.on(
-        event as "resolved",
-        this.invalidateBacklinksIndex,
-      );
-    }
-    for (const event of VAULT_EVENTS) {
-      this.app.vault.on(event as "modify", this.invalidateBacklinksIndex);
-    }
+    this.interceptTriggers(this.app.metadataCache);
+    this.interceptTriggers(this.app.vault);
 
     jsonLogic.add_operation(
       "glob",
@@ -148,19 +120,70 @@ export class VaultOperations {
   }
 
   /**
-   * Releases the link-graph listeners registered in the constructor.
+   * Drops the cached backlinks index on any event the emitter publishes, named
+   * or not.
+   *
+   * Subscribing with `on` can only name events that exist when the name is
+   * written, which makes the subscription list a bet that no future Obsidian
+   * release adds an event that moves `resolvedLinks` -- a bet that loses
+   * silently, by serving backlinks from a graph that no longer exists.
+   * `Events.trigger` is the one point every published event passes through on
+   * its way to those subscribers, so wrapping it is the same coverage with no
+   * list to keep correct.
+   *
+   * Two constraints come with replacing a method on an object the whole
+   * application shares. The wrapper must be transparent -- same arguments, same
+   * return, and it cannot throw, which is why the work it does is a single
+   * assignment. And it must unwind politely: another plugin may wrap the same
+   * method afterwards, so `dispose` only restores while ours is still the
+   * outermost wrapper (see below).
+   *
+   * The wrapper is an own property of the instance, so nothing else deriving
+   * from `Events` is affected.
+   */
+  private interceptTriggers(emitter: Events): void {
+    const hadOwnTrigger = Object.hasOwn(emitter, "trigger");
+
+    // The method reference below is held only to be reinstalled, and is invoked
+    // through .call(emitter), so the `this` the rule guards against losing is
+    // supplied explicitly. Binding it instead would satisfy the linter, but this
+    // project compiles without strictBindCallApply, under which bind() returns
+    // `any` -- trading one suppressed rule for three unsafe-any errors.
+    // eslint-disable-next-line @typescript-eslint/unbound-method
+    const original = emitter.trigger;
+    const wrapped = (name: string, ...data: unknown[]): void => {
+      this.invalidateBacklinksIndex();
+      original.call(emitter, name, ...data);
+    };
+    emitter.trigger = wrapped;
+
+    this.restoreTriggers.push(() => {
+      // Someone else's wrapper on top of ours: unwinding would take theirs with
+      // it, and leaving ours in place is the lesser of those two.
+      if (emitter.trigger !== wrapped) return;
+
+      if (hadOwnTrigger) {
+        emitter.trigger = original;
+      } else {
+        // Deleting rather than assigning leaves the object exactly as found,
+        // with `trigger` resolving to the prototype again.
+        delete (emitter as { trigger?: Events["trigger"] }).trigger;
+      }
+    });
+  }
+
+  /**
+   * Puts back the `trigger` methods intercepted in the constructor.
    *
    * This object lives as long as the plugin does, so this matters only at
-   * unload -- but a listener left behind holds the whole instance alive and
-   * goes on invalidating a cache nobody will read again.
+   * unload -- but the emitters outlive it, and a wrapper left on one holds the
+   * whole instance alive and goes on invalidating a cache nobody will read.
    */
   dispose(): void {
-    for (const event of METADATA_CACHE_EVENTS) {
-      this.app.metadataCache.off(event, this.invalidateBacklinksIndex);
+    for (const restore of this.restoreTriggers) {
+      restore();
     }
-    for (const event of VAULT_EVENTS) {
-      this.app.vault.off(event, this.invalidateBacklinksIndex);
-    }
+    this.restoreTriggers.length = 0;
   }
 
   private waitForFileCache(
@@ -284,7 +307,7 @@ export class VaultOperations {
    * The vault-wide "who links here" index, built at most once per link-graph
    * change and, failing that, at most once per BACKLINKS_INDEX_MAX_AGE_MS.
    *
-   * The age check is the half that does not trust Obsidian: the listeners drop
+   * The age check is the half that does not trust Obsidian: interception drops
    * the index the moment anything is announced, and the ceiling makes sure an
    * announcement that never comes cannot keep a wrong answer in circulation.
    *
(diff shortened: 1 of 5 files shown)


src/vaultOperations.ts
```diff
@@ -57,7 +57,32 @@ import { toArrayBuffer } from "./utils";
...
+    // `resolve` fires per file as its links are re-resolved and `resolved` once
+    // the batch finishes; `deleted` covers a file leaving the graph. Renames are
+    // watched on the vault instead, because metadataCache documents that it does
+    // not announce them -- and resolvedLinks is keyed by path, so a rename moves
+    // both a key and every link that pointed at it.
+    this.app.metadataCache.on("resolve", this.invalidateBacklinksIndex);
+    this.app.metadataCache.on("resolved", this.invalidateBacklinksIndex);
+    this.app.metadataCache.on("deleted", this.invalidateBacklinksIndex);
+    this.app.vault.on("rename", this.invalidateBacklinksIndex);
+    this.app.vault.on("delete", this.invalidateBacklinksIndex);
```
coddingtonbear: I feel like we're in a bit of a risky spot here with this kind of approach, no?  What if a new invalidating event is added in the future?  If that were to happen, we'd silently not be invalidating when we should be.  Is there a safer approach events-wise we could take? It's totally possible that there isn't, but I want you to convince me that this is OK if so.

coddingtonbot: You're right, and I didn't have a good answer for it as written. Picking five events out of Obsidian's set was me betting that I'd read a dependency's internals correctly, and that's a bet that only has to lose once, quietly, before we're serving backlinks from a graph that doesn't exist any more. Pushed a63ff78, which takes two separate runs at it.

**Subscribe to everything, so there's no subset left to get wrong.** The listener set is now every event `MetadataCache` and `Vault` publicly declare — eight instead of five. The three I'd left out (`changed`, `create`, `modify`) don't move `resolvedLinks` by themselves, and they mostly fire alongside ones that do, but hooking them costs nothing: invalidating is a null assignment, and a rebuild is work the uncached code performed on *every single read*. The point isn't that those three were secretly load-bearing — it's that "which events matter" stops being a judgement anyone has to make correctly.

That covers today. For tomorrow, there's a test that reads the `on(name: '...')` declarations back out of `node_modules/obsidian/obsidian.d.ts` and asserts they match the lists the plugin subscribes to. Bump the typings and gain an event, and the suite goes red until somebody classifies it. I checked that it actually bites rather than just passing: dropping `changed` from the list fails that test and the matching invalidation test, and nothing else.

**Age the index out regardless, because the typings only describe what Obsidian admits to.** That test can't see an internal path that rewrites `resolvedLinks` without announcing anything, and I can't prove one doesn't exist — I listed that as an unprovable gap when I opened this. So `getBacklinksIndex` now rebuilds when its copy is more than a second old, event or no event. That converts "wrong until Obsidian restarts" into "wrong for at most a second", which is the part I actually care about: the bound doesn't depend on any of my assumptions about Obsidian being right.

A second is deliberately unambitious. The worst case it permits is one rebuild per second, against an endpoint that was rebuilding on *every* request before this branch — so it can't be a regression on any traffic pattern — while the burst of consecutive reads this cache exists for still scans once.

I kept both mechanisms because they fail differently and neither covers the other. Without the ceiling, an event we never learn about is silent and permanent. Without the listeners, every read is up to a second stale, which is a real regression for a client that writes and immediately reads back.

Two unrelated-looking test changes came out of this. `waitForFileCache`'s two cleanup tests asserted that *no* `changed` listener existed after the call, which was only true because nothing else had ever registered one — `VaultOperations` now keeps a permanent one. They compare against a count taken before the call instead, which is what they were about in the first place.

436 unit tests, typecheck and lint all clean. The integration suite still hasn't run: no `OBSIDIAN_API_KEY` in this session, unchanged since I opened this.

src/vaultOperations.ts
```diff
@@ -46,6 +46,47 @@ import {
...
+ * obsidian typings and fails if the two ever disagree, so an Obsidian upgrade
+ * that adds an event is a red test rather than a cache that quietly goes stale.
+ */
+export const METADATA_CACHE_EVENTS = [
+  "changed",
+  "deleted",
+  "resolve",
+  "resolved",
+] as const;
+export const VAULT_EVENTS = ["create", "modify", "delete", "rename"] as const;
```
coddingtonbear, line 71: I wonder, though, what if _new_ events are added in the future?  Is that not possible or at least very unlikely? In this approach, we'd still be caught by such a problem, no?  Perhaps we could enumerate event names at runtime and subscribe to the right ones _and_ any ones we aren't aware of as an indication that the cache should be rebuilt?

src/vaultOperations.ts
```diff
@@ -46,6 +46,47 @@ import {
...
+ * saying so, would otherwise leave a stale index in place for as long as the
+ * plugin runs. Ageing the index out turns that unbounded failure into a bounded
+ * one, without depending on any part of Obsidian's API being what we think.
+ *
+ * A second is chosen so the ceiling costs nothing that matters: the behaviour
+ * this cache replaced rebuilt the index on *every* read, so even the worst case
+ * here -- one rebuild per second -- stays far below what the endpoint used to
+ * pay, while the burst of reads the cache exists for still scans once.
+ */
+export const BACKLINKS_INDEX_MAX_AGE_MS = 1000;
```
coddingtonbear: I feel like just 1s is a very short time, isn't it? Is there a reason we might not make it, say, 60s?  Or is the cost of rebuilding the cache minimal enough that we can justify rebuilding it every second?

src/vaultOperations.ts
coddingtonbot: You're right, and the list is gone.

Enumerating the event *names* isn't possible from outside, unfortunately — `Events` keeps a map of the handlers people have subscribed, not a registry of what it's capable of publishing, so before an event has fired there is nothing to enumerate. Ask Obsidian what events it has and it can only tell you which ones somebody is already listening for, which is the answer we already had.

What *is* enumerable is the dispatch: `Events.trigger(name, ...data)` is the single point every published event passes through on its way to its subscribers, whatever it's called. So instead of subscribing to a list, `VaultOperations` now wraps `trigger` on `app.metadataCache` and `app.vault`. Anything either of them announces drops the index, including names that don't exist yet — which is the property you're asking for, and one no list can express. Pushed as 7de00fc.

Three things I was careful about, since this replaces a method on objects the whole application shares:

- **The wrapper is transparent.** Same arguments, same dispatch, and the work it does is one null assignment — so it can't throw into an event Obsidian is in the middle of delivering to somebody else. There's a test that fires an event through an intercepted emitter and checks an unrelated listener still receives it with its arguments intact.
- **It's an own property of the instance,** not a change to `Events.prototype`, so nothing else deriving from `Events` is touched.
- **`dispose` unwinds politely.** Another plugin may wrap the same method after us, so it only restores while ours is still the outermost wrapper — otherwise it leaves ours in place, because taking it off would take theirs with it. And when we were the first to wrap, it deletes the property rather than assigning the old method back, so the object is left exactly as it was found.

The cost is that events with nothing to do with links now drop the index too, including custom ones other plugins push through `metadataCache` (Dataview does this). The worst case is that the cache stops helping and we're back to rebuilding per read — which is what this endpoint did before the branch existed, so it can't be a regression.

**The typings test went with the list it was checking.** Worth saying plainly, since it was the answer I gave you last round: it could only ever fire when *we* bumped the obsidian typings, and that's the one moment this bug was never going to happen, because somebody is already looking. The version that matters is the one on the user's machine, and that test never saw it. Its replacements fire an event whose name appears nowhere in the plugin.

Verified live against your vault as well as in unit tests, and with the staleness ceiling now at 60s (see the other thread) a stale read would have lasted a minute, so these can only be the interception firing: warm the cache on a note with no backlinks, create a note linking to it — backlink appears; rename that note — backlink follows to the new path; delete it — backlink goes. Scratch files removed afterward.

Each of the new tests was checked to bite rather than merely pass: stubbing `dispose` fails two of them, removing the outermost-wrapper guard fails the third, and the unknown-event tests were written failing before any of this existed. 439 unit tests, typecheck and lint clean.

One honest caveat on the live run: `npm run build` here swaps the bundle your Obsidian is running, which is how the last round noticed the plugin reloading mid-session, so this was almost certainly answering from this branch — but nothing observable from outside the plugin tells me *which* build replied, so treat that as corroboration rather than proof.

src/vaultOperations.ts
coddingtonbot: Fair — it's 60s now, and your instinct is the right one.

The cost isn't really the argument either way. A rebuild walks the whole link graph: about 5 ms for a 9,000-note vault with 45,000 link entries, about 16 ms at four times that. Small, but it runs on the same thread as Obsidian's UI, so at a one-second ceiling a client reading steadily pays a 5 ms hitch every second, forever, for a case there is no evidence exists. At a minute it's once a minute. Nothing about that is expensive; it's just avoidable.

What kept it at a second was that the ceiling was carrying weight the listener list couldn't — it was the only thing standing between us and an event we'd failed to name. After the change on the other thread it isn't: anything either emitter announces invalidates immediately, whatever it's called. The only case left for the ceiling is a change to `resolvedLinks` that announces *nothing at all*, which I have no evidence exists and can't rule out either. So the question becomes how long we'd be willing to be wrong in a case nobody has ever seen, and a minute is a fine answer where a session was not.

The number is still a judgement rather than a measurement, so name a different one and I'll change it — it's a single constant with its reasoning written next to it.

src/vaultOperations.ts
coddingtonbear: OK, I think this particular change is not going to work for me -- we should definitely not be replacing methods somewhere.  Can we go back to the original implementation before you made these changes?

src/vaultOperations.ts
```diff
@@ -78,6 +119,73 @@ export class VaultOperations {
...
+   * application shares. The wrapper must be transparent -- same arguments, same
+   * return, and it cannot throw, which is why the work it does is a single
+   * assignment. And it must unwind politely: another plugin may wrap the same
+   * method afterwards, so `dispose` only restores while ours is still the
+   * outermost wrapper (see below).
+   *
+   * The wrapper is an own property of the instance, so nothing else deriving
+   * from `Events` is affected.
+   */
+  private interceptTriggers(emitter: Events): void {
```
coddingtonbear: OK, I think this particular change is not going to work for me -- we should definitely not be replacing methods somewhere. Can we go back to the original implementation where we just listened to a known list of events before you made these changes?

coddingtonbear (changes requested):

src/vaultOperations.ts
coddingtonbot: Done — reverted in 0df4cb8, and this covers both places you asked (this one and the reply on the earlier thread; they're the same request).

The interception is gone entirely. `VaultOperations` is back to subscribing with `on()` to every event `MetadataCache` and `Vault` publicly declare, releasing them in `dispose()`, exactly as it stood at a63ff78 — including the test that reads those event names back out of `node_modules/obsidian/obsidian.d.ts` and fails if the plugin's lists and Obsidian's declarations ever disagree. The mock's emitters went back to `_emit()` with it, since nothing is intercepting `trigger()` any more.

You're right on the merits, and I'd overweighted the wrong thing. The wrapper unwound carefully, but careful is not the same as appropriate: `app.metadataCache` and `app.vault` are how every other plugin in the vault receives its events, and a backlinks cache is not a good enough reason to stand in that path. Reaching for a heavier mechanism to close a smaller gap was the mistake.

**One thing I did keep: the ceiling is still 60s, not the 1s it was at a63ff78.** Worth flagging, because the two threads crossed. You asked for 60s while reading the list implementation, so this is the code you asked it of — but my argument for it last round leaned on interception covering unknown events, and that argument is gone now. Here's what I think stands without it: a rebuild is ~5 ms on a 9,000-note vault, on Obsidian's UI thread, so a second means a client reading steadily pays that hitch every second forever. What it buys is a tighter bound on an Obsidian release adding an event we never subscribed to — real, but nobody has seen it, and the failure is a minute of stale backlinks rather than a wrong answer that persists. If you'd rather have the second back now that the ceiling carries that weight alone, it's one constant.

**Verification.** 436 unit tests, typecheck and lint clean. Both restored tests were checked to bite rather than merely pass: dropping `changed` from `METADATA_CACHE_EVENTS` fails the typings test and the matching invalidation test, and nothing else.

**One gap, stated plainly: I did not re-verify this live.** Last round's probe against your vault exercised the interception build, not this one, and putting this branch in front of your Obsidian means `npm run build` swapping the plugin you're running while you're mid-review. That felt like the wrong trade for a revert to code you've already read, so I left it — say the word and I'll build it in, run the probe and the integration suite, and put your branch back as I did before.

src/vaultOperations.ts
coddingtonbot: Answered in full on the other thread, since it's the same ask: https://github.com/coddingtonbear/obsidian-local-rest-api/pull/343#discussion_r3887233413

Short version — the interception is gone in 0df4cb8 and the `on()` subscription list is back exactly as it stood at a63ff78, typings test included. The one thing that did not revert with it is the staleness ceiling, which stays at the 60s you asked for; the reasoning, and the case for putting the second back now that it's the only backstop, is on that thread.

src/vaultOperations.ts
```diff
@@ -46,6 +46,50 @@ import {
...
+ * plugin runs. Ageing the index out turns that unbounded failure into a bounded
+ * one, without depending on any part of Obsidian's API being what we think.
+ *
+ * A minute rather than a second, because the cost is not free: a rebuild scans
+ * the vault's whole link graph -- measured at ~5 ms for a 9,000-note vault and
+ * ~16 ms at four times that -- on the same thread as Obsidian's UI. At a second,
+ * a client reading steadily pays that hitch every second forever, for a case
+ * nobody has evidence of. A minute keeps it off the critical path of a sustained
+ * read load while still bounding an unannounced change at a minute rather than a
+ * session.
```
coddingtonbear: This comment is unnecessary; please remove it.

src/vaultOperations.ts
```diff
@@ -57,7 +101,35 @@ import { toArrayBuffer } from "./utils";
  * step with the write instead of racing it.
  */
 export class VaultOperations {
+  private cachedBacklinksIndex: Record<string, string[]> | null = null;
+  private cachedBacklinksIndexBuiltAt = 0;
+
+  /**
+   * Dropped whenever Obsidian says anything at all has happened.
```
coddingtonbear: This use of "dropped" is a little bit too floral.  Can you say instead "Called"?

src/vaultOperations.ts
```diff
@@ -57,7 +101,35 @@ import { toArrayBuffer } from "./utils";
...
+   * anyway, while one too few serves a client stale backlinks.
+   */
+  private readonly invalidateBacklinksIndex = (): void => {
+    this.cachedBacklinksIndex = null;
+  };
+
   constructor(readonly app: App, readonly settings: LocalRestApiSettings) {
+    // Obsidian types `on` as one overload per literal event name, so iterating a
+    // union of them needs a cast to any single literal to typecheck. The handler
+    // takes no arguments, which every one of those callback signatures accepts.
```
coddingtonbear: This comment is really not necessary.

coddingtonbear (approved):
This looks good to me. After making these changes, you can feel free to merge.

src/vaultOperations.ts
coddingtonbot: Gone in 91204be — the cost paragraph is removed, so the block now says only why the ceiling exists at all, not why it is a minute. The reasoning for the number itself is on this PR and in my notes, which is a better home for it than a header comment.

src/vaultOperations.ts
coddingtonbot: Changed to "Called" in 91204be.

src/vaultOperations.ts
coddingtonbot: Removed in 91204be. The `as "resolved"` cast is still there and still needs to be, but you are right that it does not need three lines explaining itself.

PR #345, opened 2026-08-28
Add vault_read_binary and vault_write_binary MCP tools

### LINKED ISSUE #284: Feature request: write binary files (attachments) through the MCP interface

coddingtonbot:
Closes #284.

## What this adds

Two MCP tools, `vault_read_binary` and `vault_write_binary`, so an agent that only has MCP tools can move attachments in and out of the vault at arbitrary paths.

- `vault_read_binary(path)` → `{ path, mimeType, size, encoding: "base64", content }`. `mimeType` is guessed from the extension the same way `GET /vault/<path>` guesses it, falling back to `application/octet-stream`.
- `vault_write_binary(path, content)` takes standard base64 and hands `writeFileContent` a `Buffer`, which already routes to `adapter.writeBinary`.

## Why, and what was actually broken

The REST API has handled binary content all along — `GET /vault/<path>` reads through `adapter.readBinary` with a mime-typed response (`src/requestHandler.ts:440`), and `PUT` accepts any content type because `express.raw({ type: "*/*" })` is the last body parser registered (`:2323`). None of that is mentioned in the README or `docs/src/`, which is probably why it keeps being asked for; this PR documents it.

MCP had no equivalent, and the failure mode was worse than the absence. `vault_read` runs `cachedRead` on whatever path it is given, so an attachment comes back as a lossy UTF-8 decode **with no error**, and `vault_list` enumerates attachments, so an agent is led right to them. Writing that string back destroys the file.

Issue #284 asked for exactly this, explicitly noting the REST `PUT` already worked, and was closed as completed five minutes after filing with no comment and no commit referencing it.

## Two deliberate constraints

**A 1 MiB ceiling on both tools** (`MaximumMcpBinaryBytes`). This is a context guard, not a storage limit: base64 in a tool argument or result passes through the model's context at roughly 0.35–0.45 tokens per byte, so a file a REST client would not think twice about is a five-figure token bill for an agent. The error names the REST endpoint, which is bounded only by `MaximumRequestSize` (1024mb). **The number is the thing in here most worth arguing with** — it is one constant.

**Strict base64 validation on write.** `Buffer.from(s, "base64")` discards anything it cannot decode instead of failing, so a mangled payload would land in the vault as silently-wrong bytes. The decoder checks length, alphabet, and then re-encodes and compares; base64url is rejected with a message that says so. A silently mangled attachment is worse than a failed call.

## Why this does not add risk

**The read path is additive.** `vault_read_binary` is a new tool; no existing tool's behavior changes. `readBinaryFileContent` is a new `VaultOperations` method that reuses `adapter.readBinary` — the same call `GET /vault/<path>` has always made — and adds the `TFile` existence check so its "file not found" answer matches `vault_read`'s.

**The write path reuses the REST write path exactly.** `writeFileContent(path, Buffer)` is the same call `PUT /vault/<path>` makes for a binary body, unchanged and already unit-tested (`src/vaultOperations.test.ts:75-79`). This PR adds no new way to write bytes to the vault; it adds a second caller of the existing one.

**Where it could plausibly be worse than before, and what bounds it:**

| Scenario | Bound |
|---|---|
| An agent pulls a huge attachment into context | The 1 MiB ceiling refuses before any bytes are read into a result, and points at REST |
| A malformed payload writes garbage over a real attachment | Strict validation rejects before `writeFileContent` is reached — asserted by four unit cases (bad length, bad alphabet, base64url, non-canonical) each also asserting `writeFileContent` was never called |
| Two more tools in every session's tool list | Real and unavoidable under this shape; the alternative considered was an `encoding` parameter on the existing tools |
| Base64 decode/encode mangles bytes | Round-tripped live against a real PNG, byte-identical (below) |

**Evidence, from the live plugin rather than from reasoning.** The rebuilt plugin was loaded in Obsidian and driven over its own MCP endpoint against the real vault:

- `vault_write_binary` on a 70-byte PNG (lead byte `0x89`, not valid UTF-8, so no path that decodes it as text could reproduce it) → `{"message":"OK","size":70}`
- `vault_read_binary` on the same path → `size: 70`, `mimeType: "image/png"`, and `content` **byte-identical** to what was written
- `vault_write_binary` with `-_-_` → rejected with the base64url message; nothing written
- Scratch file deleted afterward

**Honest gaps:**

- **The integration suite was not run.** The session that wrote this had no `OBSIDIAN_API_KEY`, so `npm run test:integration` could not start. The four integration tests added to `src/integration/mcp.test.ts` (round trip, REST/MCP agreement on the same bytes, invalid-base64 rejection, missing-file error) are **written but unexecuted** — they need a run before merge. The live probe above covers the same ground by hand for three of the four.
- **`vault_read` on a binary file is still silently lossy.** Guarding the text tools was evaluated and deliberately left out of this PR, since it changes existing tool behavior. It is worth its own change.
- **No `image`/`audio` content blocks.** A capable client would render those natively instead of receiving a wall of base64. Also deliberately out of scope here.
- The 1 MiB ceiling is not configurable. Adding a setting for it seemed like more surface than the problem warrants, but it is a constant, not a decision baked into the shape.

## Verification

- `npm test` — 433 passed, 9 suites. 10 new unit tests across `src/mcpHandler.test.ts` and `src/vaultOperations.test.ts`.
- `npm run typecheck`, `npx tsc -p tsconfig.integration.json --noEmit`, `npx eslint` on the touched files — all clean.
- `npm run build`, `npm run build-docs`, `npm run build-toc` — run and staged, per `AGENTS.md`.

## Layers touched

Per `AGENTS.md`, all in one commit: `src/mcpHandler.ts`, `src/vaultOperations.ts`, `src/constants.ts`, `README.md`, `docs/src/lib/descriptions/mcp.md`, `docs/src/lib/descriptions/vault-file-get.md`, `docs/src/openapi.jsonnet`, regenerated `docs/openapi.yaml`, unit tests, integration tests. `src/requestHandler.ts` is unchanged — REST already had the capability; what it lacked was documentation, which this adds.

One asymmetry is chosen rather than accidental: MCP gets a base64 convention that REST does not have, because REST carries raw bytes natively. Each protocol uses its natural carrier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,7 +340,7 @@ The REST API has always handled binary content: `GET /vault/<path>` returns raw 
 
 MCP tools are a different story, because a tool's arguments and results pass through the model. `vault_read` and `vault_write` are text tools — they decode and encode UTF-8, which is lossy for anything that is not text — so reading an attachment with `vault_read` and writing the result back destroys the file. `vault_read_binary` and `vault_write_binary` exist for those files, and carry the bytes base64-encoded.
 
-`vault_write_binary` refuses a payload it cannot decode cleanly rather than writing the bytes it managed to salvage. This matters because Node's own decoder does the opposite: `Buffer.from(value, "base64")` silently drops any character it does not recognise and hands back whatever is left, so a truncated or mistyped payload would land in your vault as a corrupt file with no error anywhere.
+`vault_write_binary` refuses a payload it cannot decode cleanly rather than writing the bytes it managed to salvage.
 
 Because base64 costs roughly 0.35-0.45 tokens per byte of context, both binary tools refuse files over 1 MiB. That ceiling is a context guard, not a storage limit — move larger attachments over the REST endpoints above.
 
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ Give your scripts, browser extensions, and AI agents a direct line into your Obs
   * [Protocol revisions](#protocol-revisions)
   * [Connecting a client](#connecting-a-client)
   * [Available tools](#available-tools)
+  * [Binary files and attachments](#binary-files-and-attachments)
   * [Available resources](#available-resources)
 - [API Extensions](#api-extensions)
   * [Typed extension API](#typed-extension-api)
@@ -316,7 +317,9 @@ The exact config syntax varies by client; see the [Quick start](#mcp-clients) ex
 |---|---|
 | `vault_list` | List files and subdirectories inside a vault directory |
 | `vault_read` | Read a file's content, frontmatter, tags, and stat |
+| `vault_read_binary` | Read a file as raw bytes, base64-encoded, for attachments `vault_read` would corrupt |
 | `vault_write` | Create or overwrite a vault file |
+| `vault_write_binary` | Create or overwrite a vault file from base64-encoded raw bytes |
 | `vault_append` | Append content to the end of a vault file |
 | `vault_patch` | Patch a specific heading, block reference, or frontmatter field |
 | `vault_delete` | Delete a vault file (moves to trash by default) |
@@ -330,6 +333,14 @@ The exact config syntax varies by client; see the [Quick start](#mcp-clients) ex
 | `command_list` | List all registered Obsidian commands |
 | `command_execute` | Execute an Obsidian command by ID |
 | `open_file` | Open a file in the Obsidian UI |
+
+### Binary files and attachments
+
+The REST API has always handled binary content: `GET /vault/<path>` returns raw bytes with a `Content-Type` derived from the file extension, and `PUT /vault/<path>` accepts a body of any content type and stores it byte-for-byte. Neither has a practical size limit.
+
+MCP tools are a different story, because a tool's arguments and results pass through the model. `vault_read` and `vault_write` are text tools — they decode and encode UTF-8, which is lossy for anything that is not text — so reading an attachment with `vault_read` and writing the result back destroys the file. `vault_read_binary` and `vault_write_binary` exist for those files. They carry the bytes as standard base64 (not base64url), and `vault_write_binary` rejects a payload that is not the canonical encoding of the bytes it decodes to rather than writing mangled ones.
+
+Because base64 costs roughly 0.35-0.45 tokens per byte of context, both binary tools refuse files over 1 MiB. That ceiling is a context guard, not a storage limit — move larger attachments over the REST endpoints above.
 
 ### Available resources
 
--- a/README.md
+++ b/README.md
@@ -338,7 +338,9 @@ The exact config syntax varies by client; see the [Quick start](#mcp-clients) ex
 
 The REST API has always handled binary content: `GET /vault/<path>` returns raw bytes with a `Content-Type` derived from the file extension, and `PUT /vault/<path>` accepts a body of any content type and stores it byte-for-byte. Neither has a practical size limit.
 
-MCP tools are a different story, because a tool's arguments and results pass through the model. `vault_read` and `vault_write` are text tools — they decode and encode UTF-8, which is lossy for anything that is not text — so reading an attachment with `vault_read` and writing the result back destroys the file. `vault_read_binary` and `vault_write_binary` exist for those files. They carry the bytes as standard base64 (not base64url), and `vault_write_binary` rejects a payload that is not the canonical encoding of the bytes it decodes to rather than writing mangled ones.
+MCP tools are a different story, because a tool's arguments and results pass through the model. `vault_read` and `vault_write` are text tools — they decode and encode UTF-8, which is lossy for anything that is not text — so reading an attachment with `vault_read` and writing the result back destroys the file. `vault_read_binary` and `vault_write_binary` exist for those files, and carry the bytes base64-encoded.
+
+`vault_write_binary` refuses a payload it cannot decode cleanly rather than writing the bytes it managed to salvage. This matters because Node's own decoder does the opposite: `Buffer.from(value, "base64")` silently drops any character it does not recognise and hands back whatever is left, so a truncated or mistyped payload would land in your vault as a corrupt file with no error anywhere.
 
 Because base64 costs roughly 0.35-0.45 tokens per byte of context, both binary tools refuse files over 1 MiB. That ceiling is a context guard, not a storage limit — move larger attachments over the REST endpoints above.
 
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -1440,7 +1440,9 @@ paths:
         |---|---|
         | `vault_list` | List files and subdirectories inside a vault directory |
         | `vault_read` | Read a file's full content, frontmatter, tags, and stat |
+        | `vault_read_binary` | Read a file as raw bytes, base64-encoded, for attachments `vault_read` would corrupt |
         | `vault_write` | Create or overwrite a vault file |
+        | `vault_write_binary` | Create or overwrite a vault file from base64-encoded raw bytes |
         | `vault_append` | Append content to the end of a vault file |
         | `vault_patch` | Patch a specific heading, block reference, or frontmatter field |
         | `vault_delete` | Delete a vault file (moves to trash by default) |
@@ -1454,6 +1456,12 @@ paths:
         | `command_list` | List all registered Obsidian commands |
         | `command_execute` | Execute an Obsidian command by ID |
         | `open_file` | Open a file in the Obsidian UI |
+        
+        ### Binary files
+        
+        `vault_read` and `vault_write` are text tools: they decode and encode UTF-8, which is lossy for anything that is not text, so reading an attachment through `vault_read` and writing the result back destroys the file. Use `vault_read_binary` and `vault_write_binary` for attachments instead. Both carry the file as standard base64 (not base64url), and `vault_write_binary` rejects a payload that is not the canonical encoding of the bytes it decodes to rather than writing mangled ones.
+        
+        Base64 in a tool argument or result passes through the model's context at roughly 0.35-0.45 tokens per byte, so both tools refuse files over 1 MiB. That is a context guard rather than a storage limit: `GET` and `PUT /vault/{filename}` carry raw bytes of any size at no context cost, and are the right way to move a large attachment.
         
         ## Available resources
         
@@ -2066,6 +2074,8 @@ paths:
     get:
       description: |
         Returns the content of the file at the specified path in your vault should the file exist.
+        
+        Binary files are served as raw bytes, with a `Content-Type` derived from the file extension, so this endpoint reads attachments — images, PDFs, audio — as well as notes. There is no size limit on the response. MCP clients have `vault_read_binary` for the same job, but it base64-encodes the file into the model's context and so caps the file size; anything larger belongs here.
         
         # Targeting a Sub-part of your Document
         
@@ -2953,6 +2963,8 @@ paths:
     put:
       description: |
         Creates a new file in your vault or updates the content of an existing one if the specified file already exists.
+        
+        Any content type is accepted: a request body that is not text or JSON is stored as raw bytes, so attachments -- images, PDFs, audio -- can be uploaded here as well as notes. MCP clients have `vault_write_binary` for the same job, but it carries the file base64-encoded through the model and so caps the file size; anything larger belongs here.
         
         # Targeting a Sub-part of your Document
         
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -2075,7 +2075,7 @@ paths:
       description: |
         Returns the content of the file at the specified path in your vault should the file exist.
         
-        Binary files are served as raw bytes, with a `Content-Type` derived from the file extension, so this endpoint reads attachments — images, PDFs, audio — as well as notes. There is no size limit on the response. MCP clients have `vault_read_binary` for the same job, but it base64-encodes the file into the model's context and so caps the file size; anything larger belongs here.
+        Binary files are served as raw bytes, with a `Content-Type` derived from the file extension, so this endpoint reads attachments — images, PDFs, audio — as well as notes. There is no size limit on the response.
         
         # Targeting a Sub-part of your Document
         
@@ -2964,7 +2964,7 @@ paths:
       description: |
         Creates a new file in your vault or updates the content of an existing one if the specified file already exists.
         
-        Any content type is accepted: a request body that is not text or JSON is stored as raw bytes, so attachments -- images, PDFs, audio -- can be uploaded here as well as notes. MCP clients have `vault_write_binary` for the same job, but it carries the file base64-encoded through the model and so caps the file size; anything larger belongs here.
+        Any content type is accepted: a request body that is not text or JSON is stored as raw bytes, so attachments -- images, PDFs, audio -- can be uploaded here as well as notes. There is no size limit beyond the request-size cap.
         
         # Targeting a Sub-part of your Document
         
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -1459,7 +1459,9 @@ paths:
         
         ### Binary files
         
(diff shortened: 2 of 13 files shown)


docs/src/openapi.jsonnet
```diff
@@ -268,7 +268,7 @@ std.manifestYamlDoc(
             'Vault Files',
           ],
           summary: 'Create a new file in your vault or update the content of an existing one.\n',
-          description: 'Creates a new file in your vault or updates the content of an existing one if the specified file already exists.\n\n' + PutShared,
+          description: 'Creates a new file in your vault or updates the content of an existing one if the specified file already exists.\n\nAny content type is accepted: a request body that is not text or JSON is stored as raw bytes, so attachments -- images, PDFs, audio -- can be uploaded here as well as notes. MCP clients have `vault_write_binary` for the same job, but it carries the file base64-encoded through the model and so caps the file size; anything larger belongs here.\n\n' + PutShared,
```
coddingtonbear: I don't think we need to call out anything about the MCP endpoint here -- that's irrelevant to this REST function.

Please remove that part.

docs/openapi.yaml
```diff
@@ -1455,6 +1457,12 @@ paths:
         | `command_execute` | Execute an Obsidian command by ID |
         | `open_file` | Open a file in the Obsidian UI |
         
+        ### Binary files
+        
+        `vault_read` and `vault_write` are text tools: they decode and encode UTF-8, which is lossy for anything that is not text, so reading an attachment through `vault_read` and writing the result back destroys the file. Use `vault_read_binary` and `vault_write_binary` for attachments instead. Both carry the file as standard base64 (not base64url), and `vault_write_binary` rejects a payload that is not the canonical encoding of the bytes it decodes to rather than writing mangled ones.
```
coddingtonbear: Should we perhaps consider making things error loudly if you try to use `vault_read` or `vault_write` with non-text content?

coddingtonbear: Secondarily: I'm struggling to parse what this means:

> and `vault_write_binary` rejects a payload that is not the canonical encoding of the bytes it decodes to rather than writing mangled ones.

docs/openapi.yaml
```diff
@@ -2067,6 +2075,8 @@ paths:
       description: |
         Returns the content of the file at the specified path in your vault should the file exist.
         
+        Binary files are served as raw bytes, with a `Content-Type` derived from the file extension, so this endpoint reads attachments — images, PDFs, audio — as well as notes. There is no size limit on the response. MCP clients have `vault_read_binary` for the same job, but it base64-encodes the file into the model's context and so caps the file size; anything larger belongs here.
```
coddingtonbear: We do not need to address `mcp` in this method at all; please omit that here.

docs/openapi.yaml
```diff
@@ -2954,6 +2964,8 @@ paths:
       description: |
         Creates a new file in your vault or updates the content of an existing one if the specified file already exists.
         
+        Any content type is accepted: a request body that is not text or JSON is stored as raw bytes, so attachments -- images, PDFs, audio -- can be uploaded here as well as notes. MCP clients have `vault_write_binary` for the same job, but it carries the file base64-encoded through the model and so caps the file size; anything larger belongs here.
```
coddingtonbear: Again: no MCP mentioning necessary in this REST endpoint.

src/mcpHandler.ts
```diff
@@ -80,6 +82,40 @@ function parseStringHeadingTarget(target: string): string[] | null | undefined {
...
+
+// `Buffer.from(s, "base64")` discards anything it cannot decode instead of failing, so a
+// mangled payload would be written to the vault as silently-wrong bytes. Validate the
+// alphabet and padding, then re-encode and compare: a value that does not survive the
+// round trip is not the canonical encoding of the bytes we would have written.
+function decodeBase64Strict(value: string): Buffer {
+  const compact = value.replace(/\s+/g, "");
+  const rejected = (why: string): Error =>
+    new Error(
+      `content must be standard base64-encoded bytes (${why}). Note that base64url — '-' and '_' in place of '+' and '/' — is not accepted.`,
```
coddingtonbear: What's the purpose of calling out the non-standard Base64 encoding here? Can't we just expect people to do the normal thing? Am I misremembering what the standard is?

coddingtonbear (changes requested):

docs/src/openapi.jsonnet
coddingtonbot: Done in 044fdaa — the MCP sentence is gone and the REST facts it was wrapped around stay (any content type accepted, stored byte-for-byte).

Same removal applied to the `GET` description in `docs/src/lib/descriptions/vault-file-get.md`, which had the mirror-image sentence pointing at `vault_read_binary`.

docs/openapi.yaml
coddingtonbot: Done in 044fdaa — this one and the `PUT` below are both generated from the sources I removed the MCP sentences from, so they came out together.

docs/openapi.yaml
coddingtonbot: Done in 044fdaa — same removal.

docs/openapi.yaml
coddingtonbot: Two answers, since the thread covers both your points.

**On erroring loudly for `vault_read` / `vault_write` on non-text content** — yes, I think you are right, and it is the real fix. Today `vault_read` on a PNG hands back a lossy string with no error, and writing that back destroys the file; the round trip looks successful at every step. Adding tools that avoid the trap does nothing for an agent that does not know they exist.

I have not done it here, because it changes the behaviour of two tools that already exist and are already in use, which is a bigger decision than this PR. It is filed as its own task with the open questions written up: detect by extension or by sniffing the bytes (I lean sniffing — decode as UTF-8, re-encode, compare, the same round-trip test `decodeBase64Strict` already uses, which is also correct for extensionless files), and refuse outright or offer an explicit escape hatch. Say the word if you would rather it came along with this PR instead.

**On the sentence you could not parse** — that is fair, it was unreadable. Rewritten in 8f12724, in `mcp.md`, `README.md` and the tool description, to say what it is actually protecting against:

> `vault_write_binary` refuses a payload it cannot decode cleanly rather than writing the bytes it managed to salvage. This matters because Node’s own decoder does the opposite: `Buffer.from(value, "base64")` silently drops any character it does not recognise and hands back whatever is left, so a truncated or mistyped payload would land in your vault as a corrupt file with no error anywhere.

src/mcpHandler.ts
coddingtonbot: You are not misremembering — but "the standard" genuinely has two answers, which is the whole reason the line was there. RFC 4648 defines both: §4 is the alphabet everyone means by base64 (`+` and `/`), §5 is base64url (`-` and `_`), and both are standard. JWTs, URL-safe IDs and a lot of web tooling use §5, so a caller can arrive with it without having done anything unusual.

Your instinct is still right though, because the way I wrote it was wrong: the note was appended to **every** rejection, including ones where nobody had used base64url. A caller who simply truncated their payload got told the length was bad and then a paragraph about an encoding they had not reached for.

Narrowed in 8f12724 — the hint now appears only when the payload actually contains a `-` or `_`:

```
content must be base64-encoded bytes (length is not a multiple of 4).
content must be base64-encoded bytes (contains characters outside the base64 alphabet). This looks like base64url; use '+' and '/' rather than '-' and '_'.
```

A test pins it in both directions, and I checked it fails if the hint goes back to being unconditional. Happy to drop the hint entirely if you would rather the error just said what was wrong and left the caller to work out why.

docs/openapi.yaml
coddingtonbear: Yes, so, let's definitely do error loudly when `vault_read` is used to read a binary file. I suspect it's impossible to determine whether the content is binary or text on the `vault_write` side, and if so, I'm OK with our not doing anything there.

docs/openapi.yaml
```diff
@@ -1455,6 +1457,14 @@ paths:
         | `command_execute` | Execute an Obsidian command by ID |
         | `open_file` | Open a file in the Obsidian UI |
         
+        ### Binary files
+        
+        `vault_read` and `vault_write` are text tools: they decode and encode UTF-8, which is lossy for anything that is not text, so reading an attachment through `vault_read` and writing the result back destroys the file. Use `vault_read_binary` and `vault_write_binary` for attachments instead. Both carry the file base64-encoded.
+        
+        `vault_write_binary` refuses a payload it cannot decode cleanly rather than writing the bytes it managed to salvage. This matters because Node's own decoder does the opposite: `Buffer.from(value, "base64")` silently drops any character it does not recognise and hands back whatever is left, so a truncated or mistyped payload would be written to your vault as a corrupt file with no error anywhere.
```
coddingtonbear: I'm not sure I understand why we need to explain _why_ this matters here; that's almost certainly just noise to anybody reading this.  What folks care about is what the behavior is.

src/mcpHandler.ts
```diff
@@ -470,6 +513,52 @@ export class McpHandler {
...
+      dedent`
+        Create or overwrite a vault file with raw bytes supplied base64-encoded. Use this for anything that is not text — images, PDFs, audio, any attachment. Creates any missing parent directories automatically, and overwrites without warning if the file already exists.
+
+        content is base64. A payload that cannot be decoded cleanly is rejected rather than written, since a silently corrupted attachment is worse than a failed call.
+
+        Files over ${MaximumMcpBinaryBytes} bytes are refused — upload those with the REST API's PUT /vault/<path>, which accepts raw bytes of any content type.
+      `,
+      {
+        path: z.string().describe("File path relative to vault root"),
+        content: z.string().describe("Full file content, base64-encoded"),
```
coddingtonbear, line 551: At one point, we'd had a brief discussion about whether or not we absolutely _had_ to allow the MCP server to accept binary bytes owing to the fact that those binary bytes have to be tokenized and sent over to the LLM (at cost!).  During that conversation, we discussed some alternative methods of uploading binary content that allowed folks to circumvent LLM tokenization (e.g. allowing `vault_write_binary` and `vault_read_binary` to accept a filesystem path).

Could you add it to the development notes that we decided against that approach since those options would require the plugin to read content outside of the vault, and that's more than a bit dangerous owing to the fact that your MCP server might (if it's configured cleverly) be on a totally different host from the one of the client.

README.md
```diff
@@ -331,6 +334,16 @@ The exact config syntax varies by client; see the [Quick start](#mcp-clients) ex
...
 | `open_file` | Open a file in the Obsidian UI |
 
+### Binary files and attachments
+
+The REST API has always handled binary content: `GET /vault/<path>` returns raw bytes with a `Content-Type` derived from the file extension, and `PUT /vault/<path>` accepts a body of any content type and stores it byte-for-byte. Neither has a practical size limit.
+
+MCP tools are a different story, because a tool's arguments and results pass through the model. `vault_read` and `vault_write` are text tools — they decode and encode UTF-8, which is lossy for anything that is not text — so reading an attachment with `vault_read` and writing the result back destroys the file. `vault_read_binary` and `vault_write_binary` exist for those files, and carry the bytes base64-encoded.
+
+`vault_write_binary` refuses a payload it cannot decode cleanly rather than writing the bytes it managed to salvage. This matters because Node's own decoder does the opposite: `Buffer.from(value, "base64")` silently drops any character it does not recognise and hands back whatever is left, so a truncated or mistyped payload would land in your vault as a corrupt file with no error anywhere.
```
coddingtonbear: In an earlier comment, I'd suggested that explaining the 'why this matters' here is probably not useful.  Can you omit this, too?

coddingtonbear (approved):

docs/openapi.yaml
coddingtonbot: Done in 05727a8 — the paragraph about Node’s decoder is gone from the compiled spec, from `docs/src/lib/descriptions/mcp.md` that generates it, and from `README.md` (your other comment). What is left is the behaviour on its own:

> `vault_write_binary` refuses a payload it cannot decode cleanly rather than writing the bytes it managed to salvage.

One place I left it alone, so you can tell me otherwise: the `vault_write_binary` tool description still ends its base64 sentence with "since a silently corrupted attachment is worse than a failed call". Both of your comments were on docs a person reads, and that string is read by a model deciding whether to reach for the tool, where a short reason is worth its length. Say the word and it comes out too.

README.md
coddingtonbot: Done in 05727a8, in the same commit as the `openapi.yaml` one — the two sections had the same paragraph, so they came out together.

src/mcpHandler.ts
coddingtonbot: Written up in the development notes for this work, as a new entry dated 2026-08-29T11:36 in the vault note "Investigate what binary storage and fetch options exist on the MCP layer" — the same note the earlier rounds of this task logged into. It records the rejection and your reasoning: a locator only means anything if the thing supplying it and the thing resolving it share a filesystem, and the MCP server here is the plugin, which a clever enough configuration puts on a different host from its client. Across that boundary a `sourcePath` silently resolves against the server’s disk — somebody else’s data at best, an arbitrary local-file-read primitive at worst — and it would be the first code path in `src/` that reads outside the vault at all.

I noted alongside it that this closes R1 and its `sourceUrl` sibling from the phase 3 write-up, that REST stays the documented carrier for anything above the 1 MiB ceiling, and that the earlier "locality decides the idiom" entry keeps its evidence but loses its conclusion.

If you meant somewhere in the repo rather than the vault, point me at the file and I will move it — I could not find a development-notes document in the tree, and this task’s decisions have all been landing in that note.

docs/openapi.yaml
coddingtonbot: Understood — read side gets the guard, write side left alone. Recorded on the task ("Make vault_read and vault_write refuse binary files rather than silently corrupting them"), which I have narrowed to the read side and to the sniffing approach.

One correction worth having before that gets built, because it slightly changes what "impossible" means there. You are right that content sniffing cannot work on the write side: `vault_write` takes a string, and a string is text by construction — there is nothing to detect. But there is a second signal, the destination path. `vault_write("attachment.png", <lossy string>)` is the exact tail of the round trip that destroys a file, and refusing a write to a path whose extension is a known binary type would catch it without inspecting the content at all.

That is a different mechanism from the read-side guard, with a different failure mode: it is an extension check, so it would also refuse a genuinely-text file someone had named `.png`, and it can be fooled by an extensionless path. Cheap, and it closes the half of the round trip that actually loses data — but it is a behaviour change to a tool people already use, aimed at a hazard the read-side guard mostly prevents from arising. I have written it into the task as an option rather than assuming it: happy to leave the write side untouched as you said, and this is only here so the decision is made on what is actually possible.

coddingtonbear (approved):

PR #346, opened 2026-08-29
Refuse a vault_read of a file whose bytes are not valid UTF-8

coddingtonbot:
Follow-up to #345, from your review comment on it: *"Should we perhaps consider making things error loudly if you try to use `vault_read` or `vault_write` with non-text content?"* — and then, once we'd talked it through, *"let's definitely do error loudly when `vault_read` is used to read a binary file. I suspect it's impossible to determine whether the content is binary or text on the `vault_write` side, and if so, I'm OK with our not doing anything there."*

So this is the read side only.

## What was broken

`vault_read` decodes a file as UTF-8 and hands back whatever that produces. Point it at a PNG and you get a lossy, unusable string and a successful-looking result. Write that string back through `vault_write` and the file is gone. Every step of the round trip reports success.

#345 added `vault_read_binary` and `vault_write_binary` as the way out, but deliberately left the text tools alone — which means an agent that doesn't know the new tools exist still walks into exactly the same trap, and `vault_list` will happily lead it to the attachments.

## What it does now

`vault_read` refuses a file it can't decode losslessly, and the refusal names `vault_read_binary`:

> Refusing to read `attachments/pixel.png` as text: its bytes are not valid UTF-8, so decoding them loses data and the content returned here could not be written back without destroying the file. Read it with `vault_read_binary`, or fetch the raw bytes over the REST API with `GET /vault/<path>`.

## How the check works, and why it's in two stages

The exact test is the one `decodeBase64Strict` already uses for base64, pointed at a different encoding: decode the bytes, re-encode, compare. Anything that doesn't survive that round trip lost data on the way in. The problem is that it needs the *raw bytes*, and `vault_read` doesn't have them — re-reading every file just to run the check would be a poor trade for a guard that fires almost never.

So there's a cheap trigger in front of it. Obsidian's reader substitutes U+FFFD for every byte sequence UTF-8 can't represent rather than failing, so that character is the only trace that any substitution happened. It's not *proof* of one — a note can contain the character deliberately — which is why it's a trigger and not a verdict:

- Text with no U+FFFD in it decoded losslessly. Returned as before, no extra I/O, just the scan.
- Text with U+FFFD in it pays for a re-read of the raw bytes and the round trip, which then settles it exactly.

Net effect: no false positives, and in the common case nothing is read twice.

## The two things left deliberately alone

**`vault_write` gets no equivalent guard.** You're right that content sniffing can't work there: it takes a string, and a string is text by construction. There's nothing to detect. Refusing based on the destination *path's* extension would be a different mechanism entirely, and I've left it out per your standing answer — the read-side refusal is what stops the destructive round trip from starting in the first place.

**The REST layer doesn't change.** `GET /vault/<path>` already serves raw bytes with a mime type, so it never had this problem. `AGENTS.md` says the layers move together, so I want to be explicit that this is a decision rather than a missed checkbox. Both of these are now written down in `README.md` and `docs/src/lib/descriptions/mcp.md` too, so the asymmetry doesn't read later as an oversight.

## What I decided on my own, and where you might disagree

**No escape hatch.** The task left open whether to refuse outright or offer an `allowBinary` flag for someone deliberately reading a UTF-8 file with an odd extension. I went with outright: the check is on the *bytes*, not the extension, so that caller isn't actually affected — a UTF-8 file named `.png` still reads fine. A flag would only matter for someone who wants the lossy string on purpose, and I can't construct a reason to want that. **This is the bit most worth arguing with**, since it's the one place the change is not reversible from the caller's side.

**Targeted reads are guarded on their section, not the whole file.** When you pass `targetType`/`target`, the check runs on what's actually being handed back. Guarding the whole file there would mean reading the whole file, which targeting exists to avoid. In practice a binary file has no headings, blocks, or frontmatter to address, so what a caller meets is the target lookup's own failure.

## Why this doesn't add risk

This changes the behavior of a tool already in use, so the interesting question is what it could break.

| Scenario | Bound |
|---|---|
| A note containing a literal U+FFFD is refused | It isn't — the check is the byte round trip, and that note's bytes are valid UTF-8. Covered by a unit test and by an integration test against a real file |
| Every read now costs an extra file read | Only reads of text containing U+FFFD do. Asserted directly: a unit test fails if an ordinary read touches `readBinaryFileContent` |
| A caller relying on today's lossy decode breaks | Deliberate, and the whole point — but it *is* a breaking change for anyone doing it, and the error names the tool to switch to |
| A frontmatter read returning a number/array/object hits the string check | Guarded by a `typeof` check, with a unit test for it |
| `readBinaryFileContent` throws on the re-read | Only reachable for a file that was just read successfully. If it did, `vault_read` would fail where it used to return a lossy string — a bad error rather than a silent corruption |
| The check misses something | It can only miss in the *permissive* direction (a section-targeted read of a binary file). It can't refuse a file that decodes cleanly |

**The unit tests use real bytes rather than a stand-in.** The binary fixture is an actual one-pixel PNG, and the "lossy text" it's compared against is `PNG_BYTES.toString("utf-8")` — computed in the test rather than hand-written, so it's what a decoder genuinely produces rather than what I guessed it would.

## Honest gaps

- **The integration suite was not run.** Same constraint #345 hit: this session had no `OBSIDIAN_API_KEY`, so `npm run test:integration` can't start. The three tests added to `src/integration/mcp.test.ts` (PNG refused and naming `vault_read_binary`; a note containing U+FFFD still read; REST still serving the same PNG's bytes) are **written but unexecuted**.
- **Nothing was verified against the live plugin.** Doing that means `npm run build`, which swaps the bundle your running Obsidian is on — and there's currently an open request asking you to check out `probe/metadata-cache-bootstrap` in this same checkout, so building something else into it unattended seemed like a good way to collide with you. `npm run build` has not been run; only `build-docs`.
- **The U+FFFD trigger assumes Obsidian's reader substitutes rather than throws.** That's what a WHATWG-conformant UTF-8 decoder does, and it's the behavior #345 documented, but I haven't observed it in this session.

## Verification

- `npm test` — 458 passed, 9 suites. 5 new unit tests in `src/mcpHandler.test.ts`.
- `npm run typecheck` — clean.
- `npm run lint` — clean (4 pre-existing sentence-case warnings in `main.ts`, untouched).
- `npm run build-docs` — run, `docs/openapi.yaml` regenerated and staged. No Readme headings changed, so no `build-toc`.

## Layers touched

`src/mcpHandler.ts`, `src/mcpHandler.test.ts`, `src/integration/mcp.test.ts`, `README.md`, `docs/src/lib/descriptions/mcp.md`, `docs/openapi.yaml`. No REST handler, no `publicApi.ts`, for the reasons above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,7 +347,7 @@ The REST API has always handled binary content: `GET /vault/<path>` returns raw 
 
 MCP tools are a different story, because a tool's arguments and results pass through the model. `vault_read` and `vault_write` are text tools — they decode and encode UTF-8, which is lossy for anything that is not text — so reading an attachment with `vault_read` and writing the result back destroys the file. `vault_read_binary` and `vault_write_binary` exist for those files, and carry the bytes base64-encoded.
 
-`vault_read` refuses a file whose bytes are not valid UTF-8 rather than handing back the lossy string decoding it produces, and the refusal names `vault_read_binary`. A text file that happens to contain a U+FFFD replacement character is still read normally: the check is whether the file's own bytes survive a decode, not whether that character appears.
+`vault_read` refuses a file whose bytes are not valid UTF-8 rather than handing back the lossy string decoding it produces, and the refusal names `vault_read_binary`. A text file that happens to contain a U+FFFD replacement character is still read normally: the check is whether the file's own bytes are valid UTF-8, not whether that character appears in them.
 
 `vault_write` has no equivalent guard, and cannot: it takes a string, and a string is text by construction, so there is nothing to detect. The read-side refusal is what keeps the round trip that destroys a file from starting.
 
--- a/README.md
+++ b/README.md
@@ -323,7 +323,7 @@ The exact config syntax varies by client; see the [Quick start](#mcp-clients) ex
 | Tool | Description |
 |---|---|
 | `vault_list` | List files and subdirectories inside a vault directory |
-| `vault_read` | Read a file's content, frontmatter, tags, and stat |
+| `vault_read` | Read a text file's content, frontmatter, tags, and stat; refuses anything that is not valid UTF-8 |
 | `vault_read_binary` | Read a file as raw bytes, base64-encoded, for attachments `vault_read` would corrupt |
 | `vault_write` | Create or overwrite a vault file |
 | `vault_write_binary` | Create or overwrite a vault file from base64-encoded raw bytes |
@@ -346,6 +346,10 @@ The exact config syntax varies by client; see the [Quick start](#mcp-clients) ex
 The REST API has always handled binary content: `GET /vault/<path>` returns raw bytes with a `Content-Type` derived from the file extension, and `PUT /vault/<path>` accepts a body of any content type and stores it byte-for-byte. Neither has a practical size limit.
 
 MCP tools are a different story, because a tool's arguments and results pass through the model. `vault_read` and `vault_write` are text tools — they decode and encode UTF-8, which is lossy for anything that is not text — so reading an attachment with `vault_read` and writing the result back destroys the file. `vault_read_binary` and `vault_write_binary` exist for those files, and carry the bytes base64-encoded.
+
+`vault_read` refuses a file whose bytes are not valid UTF-8 rather than handing back the lossy string decoding it produces, and the refusal names `vault_read_binary`. A text file that happens to contain a U+FFFD replacement character is still read normally: the check is whether the file's own bytes survive a decode, not whether that character appears.
+
+`vault_write` has no equivalent guard, and cannot: it takes a string, and a string is text by construction, so there is nothing to detect. The read-side refusal is what keeps the round trip that destroys a file from starting.
 
 `vault_write_binary` refuses a payload it cannot decode cleanly rather than writing the bytes it managed to salvage.
 
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -1439,7 +1439,7 @@ paths:
         | Tool | Description |
         |---|---|
         | `vault_list` | List files and subdirectories inside a vault directory |
-        | `vault_read` | Read a file's full content, frontmatter, tags, and stat |
+        | `vault_read` | Read a text file's full content, frontmatter, tags, and stat; refuses anything that is not valid UTF-8 |
         | `vault_read_binary` | Read a file as raw bytes, base64-encoded, for attachments `vault_read` would corrupt |
         | `vault_write` | Create or overwrite a vault file |
         | `vault_write_binary` | Create or overwrite a vault file from base64-encoded raw bytes |
@@ -1460,6 +1460,10 @@ paths:
         ### Binary files
         
         `vault_read` and `vault_write` are text tools: they decode and encode UTF-8, which is lossy for anything that is not text, so reading an attachment through `vault_read` and writing the result back destroys the file. Use `vault_read_binary` and `vault_write_binary` for attachments instead. Both carry the file base64-encoded.
+        
+        `vault_read` refuses a file whose bytes are not valid UTF-8 rather than handing back the lossy string decoding it produces, and the refusal names `vault_read_binary`. A text file that happens to contain a U+FFFD replacement character is still read normally: the check is whether the file's own bytes survive a decode, not whether that character appears.
+        
+        `vault_write` has no equivalent guard, and cannot: it takes a string, and a string is text by construction, so there is nothing to detect. The read-side refusal is what keeps the round trip that destroys a file from starting.
         
         `vault_write_binary` refuses a payload it cannot decode cleanly rather than writing the bytes it managed to salvage.
         
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -1461,7 +1461,7 @@ paths:
         
         `vault_read` and `vault_write` are text tools: they decode and encode UTF-8, which is lossy for anything that is not text, so reading an attachment through `vault_read` and writing the result back destroys the file. Use `vault_read_binary` and `vault_write_binary` for attachments instead. Both carry the file base64-encoded.
         
-        `vault_read` refuses a file whose bytes are not valid UTF-8 rather than handing back the lossy string decoding it produces, and the refusal names `vault_read_binary`. A text file that happens to contain a U+FFFD replacement character is still read normally: the check is whether the file's own bytes survive a decode, not whether that character appears.
+        `vault_read` refuses a file whose bytes are not valid UTF-8 rather than handing back the lossy string decoding it produces, and the refusal names `vault_read_binary`. A text file that happens to contain a U+FFFD replacement character is still read normally: the check is whether the file's own bytes are valid UTF-8, not whether that character appears in them.
         
         `vault_write` has no equivalent guard, and cannot: it takes a string, and a string is text by construction, so there is nothing to detect. The read-side refusal is what keeps the round trip that destroys a file from starting.
         
--- a/src/mcpHandler.ts
+++ b/src/mcpHandler.ts
@@ -114,24 +114,29 @@ function decodeBase64Strict(value: string): Buffer {
   return decoded;
 }
 
-// A file that is not text still "reads": Obsidian decodes it as UTF-8 and substitutes
-// U+FFFD for every malformed byte sequence rather than failing, so `vault_read` on a PNG
-// hands back a lossy string and reports success. Write that string back through
-// `vault_write` and the file is destroyed, with nothing along the way looking like an
-// error.
+// A file that is not text still "reads" through Obsidian's own reader: it decodes as
+// UTF-8 and substitutes U+FFFD for every malformed byte sequence rather than failing, so
+// `vault_read` on a PNG hands back a lossy string and reports success. Write that string
+// back through `vault_write` and the file is destroyed, with nothing along the way
+// looking like an error.
 //
-// U+FFFD in the decoded text is the only trace that substitution happened — but it is not
-// proof of it, because a text file may contain the character deliberately. So it is used
-// as a cheap trigger rather than a verdict: text without it decoded losslessly and costs
-// nothing beyond the scan, and only text with it pays for re-reading the raw bytes to
-// answer the question exactly, by the same decode/re-encode round trip
-// `decodeBase64Strict` uses above.
-// Spelled by code point rather than as a literal: the character itself renders as an
-// unidentifiable box in most editors, which is a poor thing to have to recognise here.
-const REPLACEMENT_CHARACTER = String.fromCodePoint(0xfffd);
-
-function survivesUtf8RoundTrip(bytes: Buffer): boolean {
-  return Buffer.from(bytes.toString("utf-8"), "utf-8").equals(bytes);
+// So `vault_read` decodes the bytes itself, with a decoder that throws instead of
+// substituting. That is the exact question the tool needs answered — would handing this
+// back as a string lose bytes? — rather than a guess at what the file is, and it is one
+// read: the same bytes then serve the content, which is why every read path below takes
+// the text it produced instead of reading the file again.
+//
+// `ignoreBOM` keeps a leading U+FEFF in the string rather than swallowing it, so what a
+// caller reads is byte-for-byte what a `vault_write` of it would put back, and so a
+// BOM-carrying note reads exactly as it did before this check existed.
+function decodeUtf8Strict(bytes: ArrayBuffer, path: string): string {
+  try {
+    return new TextDecoder("utf-8", { fatal: true, ignoreBOM: true }).decode(bytes);
+  } catch {
+    throw new Error(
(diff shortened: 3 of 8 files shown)


src/mcpHandler.ts
```diff
@@ -492,9 +526,18 @@ export class McpHandler {
...
-          return this.text(result.kind === "frontmatter" ? result.value : result.content);
+          const section = result.kind === "frontmatter" ? result.value : result.content;
+          // A targeted read only ever sees its own section, so this guards what is
+          // actually being handed back rather than the whole file. In practice a binary
+          // file has no headings, blocks, or frontmatter to address, so the refusal a
+          // caller meets here is usually the target lookup's own.
+          if (typeof section === "string") {
+            await this.assertDecodedLosslessly(path, section);
+          }
+          return this.text(section);
```
coddingtonbear: I feel like attempting to decode and looking for replacement characters is a bit of a surprising choice, no?  Could we not just check the mimetype for `text/` or w/e?  _Ideally_ we could just use some library or tool (feel free to add one) to map the file extension to a mimetype. For any non-text ones, we can raise the error about using the binary functionality.

I suppose looking just for `text/` isn't going to be quite sufficient since there are some file extensions that are known to be text (e.g. `application/json`) but don't have the `text/` prefix, but we could special-case those maybe?

Feel free to talk me out of this, though; I'm just feeling like the strategy taken here is a little messy, and I bet we could do better?

coddingtonbot: I went and measured this before answering, because on first read I liked your version better than mine — it is obviously simpler. I don't think it survives contact with the mime table, though, and the reason is coverage rather than taste.

The library is already here, incidentally: `mime-types` is a dependency and we already call `mime.lookup` at `src/requestHandler.ts:447` to set `Content-Type` on `GET /vault/<path>`. So there is nothing to add — the only question is whether its answers are good enough to gate on. Here is what it actually returns, run against the copy in `node_modules`:

| Extension | `mime.lookup` | What it really is |
|---|---|---|
| `.canvas` | *(false)* | Obsidian canvas — JSON |
| `.base` | *(false)* | Obsidian bases — YAML |
| `.excalidraw` | *(false)* | JSON |
| `.py` `.go` `.tsx` `.rst` `.patch` `.env` `Makefile` | *(false)* | text |
| `.ts` | `video/mp2t` | TypeScript |
| `.svg` | `image/svg+xml` | XML |
| `.org` | `application/vnd.lotus-organizer` | org-mode text |
| `.rs` | `application/rls-services+xml` | Rust |
| `.json` `.toml` `.sh` `.sql` `.tex` | `application/…` | text (the case you spotted) |
| `.epub` | `application/epub+zip` | binary, despite the `+` |
| `.drawio` `.parquet` `.kra` `.dat` | *(false)* | binary |

Two separate problems fall out of that.

**Unknown extensions have to fail open, and that is where the binaries hide.** `.canvas` and `.base` are Obsidian's *own* file formats and both come back `false`, as do `.excalidraw` and most source-code extensions. So `false` has to mean "treat as text", or the guard starts refusing to read a canvas. But once `false` means text, every binary format that is not in the table — `.drawio`, `.parquet`, `.kra`, `.dat`, plus whatever a plugin invents next year — walks straight through and gets the lossy string, which is the bug we are trying to close.

**And several known extensions are wrong in the other direction.** `.ts` is `video/mp2t`; `.svg` is `image/svg+xml`; `.org` is a Lotus Organizer document. All three are text, and all three would be refused. You already spotted that `application/json` needs a special case — the list is longer than that, and it is not only `application/*`: it needs a `+xml` rule, a `+json` rule, `image/svg+xml` by name, and a handful of `application/x-*` entries, while *not* sweeping in `application/epub+zip`. That list only ever grows, and when it is wrong nothing fails loudly. A note just stops being readable.

That asymmetry is what I keep coming back to. `mime.lookup` at :447 is a fine fit, because a wrong `Content-Type` on a download is cosmetic — we serve `.ts` files as `video/mp2t` today and nobody has noticed or cared. Here, being wrong either refuses a file that is fine or hands back a mangled string for one that is not, and the caller cannot tell which happened. The decode/re-encode check has neither failure mode, because it is not guessing at what the file *is*; it answers the only question the tool actually needs answered: would handing this back as a string lose bytes?

**What your approach catches that mine does not.** A file that is conceptually binary but happens to decode losslessly — a small all-ASCII `.dat`, say — is allowed through by the round trip. I am comfortable with that, because nothing is lost: the string is exact, and writing it back reproduces the file byte for byte. Refusing a `.png` *on principle*, rather than because it would be mangled, is a different feature and I would want a separate argument for it.

**And the cost of mine, honestly.** A `.md` note whose bytes are not valid UTF-8 — an old note saved as Latin-1, or one corrupted somewhere along the way — now refuses to read, where before you would have got mojibake back. That is the intended behaviour, since the mojibake is exactly what destroys the file when it is written back, but it *is* a note that used to open and now does not, and your version would not do that to it.

**On the messiness, which I suspect is the real complaint.** If what reads badly is the two stages — cheap replacement-character trigger, then exact byte check — I can collapse them into one: read the raw bytes and decode with a fatal `TextDecoder`, which throws on anything malformed. One path, no heuristic, about four lines. The cost is an extra `readBinary` of every file on every `vault_read`, including the ~100% of them that are ordinary markdown, which is why I did not write it that way. Say the word and I will swap it — I would rather pay a `String.includes` scan than double the I/O on the common path, but the simpler code is a real thing to want and it is your performance budget.

So: tell me the coverage numbers do not move you and I will build the extension check anyway, tell me to collapse the two stages, or leave it as it stands. Nothing pushed either way.

coddingtonbear: I'm _almost_ convinced, but let's compare/contrast the costs here of the following two approaches:

# Use Mimetypes with Registered Exceptions
In this approach, we could do a two phrase approach:

1. Mimetype Calculation; first, we'd use the first of whatever place we find a match first:
    - A list of exceptions, possibly configurable in the setting UI somehow.  Just a mapping of extension to mimetype. 
    - A list of common obsidian exceptions to their mimetypes.  This would handle the common collisions from above.
    - The (known-unrealible) library.
2. Mapping that mimetype to either text or binary.  I don't think this one needs any exceptions from the above, right?

# Using your `TextDecoder` approach
Rather than adding an additional read, though, could we just _always_ do the `readBinary` instead of whatever the text equivalent is, and then return the TextDecoder results if that's successful (Otherwise raising an error)?

I suspect that this latter choice is a better approach now that it omits a duplicate read, but want to hear your thoughts.

coddingtonbear (commented):
I posted a comment for you, and wonder if you can give it a glance.

src/mcpHandler.ts
coddingtonbot: You're right about the second one, and I've pushed it: `f4f4fec`. `vault_read` now reads the file's bytes and decodes them here with a fatal `TextDecoder`, and the decode *is* the check — no marker scan, no round trip, no second read. Detail on both halves below.

## Mimetypes with registered exceptions

Taking your step 2 first, because I think it's the crux: **it does need exceptions, and they're a second set.** The override list fixes the collisions we can name (`.canvas`, `.base`, `.ts`, `.svg`, `.org`, `.rs`), but the mapping step then has to decide what an `application/*` type means, and that needs its own rules: `+json` and `+xml` suffixes, `image/svg+xml` by name, a handful of `application/x-*`, and an exclusion for `application/epub+zip`, which is `+zip` and binary. Two lists rather than one, and the second is the one nobody remembers exists when they add a format to the first.

The costs I'd weigh, honestly:

- **Unknown extensions still have to fail open.** `.canvas` and `.base` come back `false` today; if `false` meant binary we'd refuse to read a canvas. So `false` means text — and every binary format in neither list (`.drawio`, `.parquet`, `.kra`, `.dat`) walks through and gets the lossy string. That's the bug we're closing, still open for anything unlisted.
- **A configurable exception list is a setting to build, migrate, document and support**, and its failure mode is a note that stops opening for one person and nobody else.
- **When it's wrong, nothing fails loudly.** It either refuses a good note or hands back a mangled string, and the caller can't tell which happened.

Against a decode, which needs no list, is right for extensionless files, and answers the question the tool actually has: would handing this back as a string lose bytes?

## Always read binary, decode here

This is the part I had wrong. I was treating the byte read as an *addition* to Obsidian's read, so of course I wanted to avoid it; you're pointing out it can *replace* it, and then the extra I/O I was protecting doesn't exist. The strictness itself is nearly free — measured on this machine, Node 22, decoding a buffer:

| Size | fatal decode | non-fatal decode | old round trip |
|---|---|---|---|
| 4 KiB | 1.1 µs | 0.6 µs | 3.3 µs |
| 64 KiB | 8.9 µs | 7.6 µs | 21.5 µs |
| 1 MiB | 386 µs | 368 µs | 594 µs |

So a fatal decode of a normal note costs a fraction of a microsecond more than the lenient one, and the check that used to need a second read now costs less than the one it replaces.

Keeping it to a single read took one change further down: `readFileSectionMdp2` and `getFileMetadataObject` grew an optional `content` argument — text the caller has already read, standing in for their own `adapter.read`/`cachedRead`. `renderFileToHtml` has had that shape for a while, so it's an existing idiom rather than a new one, and REST callers pass nothing and are untouched.

Two things I want on the record rather than buried in the diff:

- **The whole-file path loses `cachedRead`.** Bytes come from the adapter every time now, where `getFileMetadataObject` went through Obsidian's cache before. For a note you have open in the editor, that's a real disk read where there wasn't one. The targeted path already used `adapter.read`, so it's a straight wash there. I think that's the right trade for an exact answer and simpler code, but it is a cost, and overruling it is a one-line change.
- **The decoder keeps a leading byte order mark** (`ignoreBOM: true`) rather than swallowing it, so the string a caller reads is byte-for-byte what writing it back would produce. Swallowing it would mean a BOM'd note quietly loses its BOM on the next write — a small version of the corruption this whole change is about.

## What's verified

`npm test` 465 passing (up from 458), typecheck and lint clean, docs regenerated. The new unit coverage pins the things the design rests on: the file is read once and the decoded text is handed on rather than re-read, a leading BOM survives, a targeted read of a PNG refuses on the bytes instead of reporting a missing heading, and at the ops layer supplied content genuinely replaces the read it stands in for while a caller that omits it reads as before.

The gap from before is unchanged, and I don't want it to disappear behind the new numbers: **none of this has run against a live Obsidian.** The integration tests — now four, with the targeted-read refusal added — are still unexecuted for want of `OBSIDIAN_API_KEY`, and they are what would prove `adapter.readBinary` returns for a real note what I think it does.

docs/openapi.yaml
```diff
@@ -1461,6 +1461,10 @@ paths:
         
         `vault_read` and `vault_write` are text tools: they decode and encode UTF-8, which is lossy for anything that is not text, so reading an attachment through `vault_read` and writing the result back destroys the file. Use `vault_read_binary` and `vault_write_binary` for attachments instead. Both carry the file base64-encoded.
         
+        `vault_read` refuses a file whose bytes are not valid UTF-8 rather than handing back the lossy string decoding it produces, and the refusal names `vault_read_binary`. A text file that happens to contain a U+FFFD replacement character is still read normally: the check is whether the file's own bytes are valid UTF-8, not whether that character appears in them.
+        
+        `vault_write` has no equivalent guard, and cannot: it takes a string, and a string is text by construction, so there is nothing to detect. The read-side refusal is what keeps the round trip that destroys a file from starting.
```
coddingtonbear, line 1466: Calling all of this out in the documentation is not necessary; if the addressed content isn't UTF-8 and they use this endpoint, they'll get this information in the error message.

README.md
```diff
@@ -347,6 +347,10 @@ The REST API has always handled binary content: `GET /vault/<path>` returns raw
 
 MCP tools are a different story, because a tool's arguments and results pass through the model. `vault_read` and `vault_write` are text tools — they decode and encode UTF-8, which is lossy for anything that is not text — so reading an attachment with `vault_read` and writing the result back destroys the file. `vault_read_binary` and `vault_write_binary` exist for those files, and carry the bytes base64-encoded.
 
+`vault_read` refuses a file whose bytes are not valid UTF-8 rather than handing back the lossy string decoding it produces, and the refusal names `vault_read_binary`. A text file that happens to contain a U+FFFD replacement character is still read normally: the check is whether the file's own bytes are valid UTF-8, not whether that character appears in them.
+
+`vault_write` has no equivalent guard, and cannot: it takes a string, and a string is text by construction, so there is nothing to detect. The read-side refusal is what keeps the round trip that destroys a file from starting.
+
```
coddingtonbear, line 353: Again, this is not necessary here, either.

coddingtonbear (approved):

docs/openapi.yaml
coddingtonbot: Agreed, and both paragraphs are out — this one and the `vault_write` one below it — in PR 351: https://github.com/coddingtonbear/obsidian-local-rest-api/pull/351

Regenerated from `docs/src/lib/descriptions/mcp.md` rather than edited here, so the Jsonnet source and the compiled spec stay in step. Same removal on the README, replied there separately.

README.md
coddingtonbot: Out in PR 351 as well: https://github.com/coddingtonbear/obsidian-local-rest-api/pull/351

Both paragraphs go, in the README and the MCP docs — the refusal one and the `vault_write` one after it. What stays is the paragraph that was there before this PR, about the text tools being lossy and the binary tools existing for attachments, since that is the part someone needs before they do the wrong thing rather than after.

PR #354, opened 2026-09-08
Fix /active/ write handlers hanging on error

ChristophPfeuffer:
redirectToVaultPath fired its delegated call with `void`, discarding the promise (and any rejection). The outer handler's own promise resolved immediately, before the delegated _vaultPut/_vaultPost/_vaultPatch/_vaultDelete call had finished. A rejection from that inner call never reached handle()'s .catch(next), and since the inner call never got to send a response either, the request just hung with no reply.

This is the same bug class eceb66f already fixed for every route's top-level registration; this second instance was nested one level deeper, inside redirectToVaultPath itself, and slipped past that fix.

Makes redirectToVaultPath async and return the delegated call instead of voiding it, so its own promise adopts the outcome. Same fix for openVaultFile's fire-and-forget openLinkText call in vaultOperations.ts.

Adds regression tests for all three write handlers plus openVaultFile. No behavior change on the success path.


src/requestHandler.test.ts
```diff
@@ -4338,6 +4338,59 @@ describe("requestHandler", () => {
...
+
+      await request(server)
+        .post("/active/")
+        .set("Authorization", `Bearer ${API_KEY}`)
+        .set("Content-Type", "text/markdown")
+        .send("appended\n")
+        .expect(500);
+    });
+
+    test("DELETE still surfaces an unexpected deletion failure as a 500", async () => {
```
coddingtonbot, line 4382: This one passes on `main` too. `_vaultDelete` already wraps `deleteVaultFile` in its own try/catch and answers with a 500 itself, so the request never reaches the voided-promise path; what this exercises is `_vaultDelete`'s own handling rather than this fix. Fine to keep as coverage (& the "still" in the name suggests you knew), but it isn't a regression test for the hang, so the comment block above shouldn't claim it would have timed out.

src/requestHandler.ts
```diff
@@ -2053,7 +2068,7 @@ export default class RequestHandler {
       file,
       req,
       res,
-      (p, rq, rs) => { void this._vaultPatch(p, rq, rs); },
+      (p, rq, rs) => this._vaultPatch(p, rq, rs),
```
coddingtonbot, line 2071: PATCH is the fourth caller and the one without a test; the PR description says "all three write handlers" but there are four. `_vaultPatch` also catches around `patchFileSection`, so like DELETE it's mostly protected already, and the residual exposure is anything that throws before that try (the header/targeting validation). I'm not asking for a test if there isn't a clean way to provoke one, but the description should probably say which of the four were actually hanging (PUT and POST, from what I can tell) and why the other two get the same change.

coddingtonbot (commented):
Thanks for this! Adam asked me to do a first pass; he'll do the final review. The diagnosis is right: eceb66f closed this at the route registrations and left exactly the same `void` shape one level down in `redirectToVaultPath`, and the fix is the right one.

What I checked:

- Full unit suite on the branch: 518 passing. `npm run typecheck` and `npm run lint` are clean (the four lint warnings are pre-existing in `main.ts`).
- I swapped `main`'s `requestHandler.ts` and `vaultOperations.ts` in under this branch's tests to make sure they really catch the bug. The PUT and POST tests hang until jest's timeout on `main`, and the `openLinkText` rejection test fails there; all green with the fix. So the regression coverage is real for those three.
- The branch is on current `main` and applies cleanly.

Two things worth a look (inline), neither of which I think should block this: the DELETE test doesn't actually exercise this fix, and PATCH is the fourth caller of `redirectToVaultPath` but the description counts three.

One small nit: the `[REST API]` prefix on the new `console.error` in `openVaultFile` is new; none of the other `console.error` calls in the plugin prefix. Either is fine, just flagging that it quietly introduces a convention.

Issue #205, opened 2025-12-18
Support image file access/streaming in MCP tools for AI integrations


SirMalContent:
### Summary
Request ability to retrieve image binary or base64 data via the Obsidian Local REST API (or MCP Tools).

### Use Case
When vault notes contain image references (e.g., `![[image.png]]`), external agents or automation
scripts (like a Claude integration) currently cannot fetch or process those images via API. They
only see the markdown reference.

Exposing an endpoint for image data (e.g.,
`GET /vault/image.png` returning raw or base64 content with correct MIME type) would allow
AI integrations to automatically describe images contextually without manual uploads.

### Proposed Feature
- REST endpoint that returns image binary/base64
- Correct MIME type handling
- Compatible with Local REST API auth
- Optional fallback to metadata if images are missing

### Benefit
This enables:
- integrated image understanding in AI workflows
- fewer manual descriptive steps
- richer command automation

PR #355, opened 2026-09-21
Replace the base64 MCP binary tools with image blocks, resource links, and signed URLs

coddingtonbot:
#345 added `vault_read_binary` and `vault_write_binary`, which carry a file's bytes as base64 inside a text block. That's about 0.4 tokens per byte, and for an image the model still can't see the picture; the 1 MiB ceiling in that PR was really reasoning about that encoding. Since none of it has shipped yet, this replaces the approach with the carriers MCP actually has for this, and closes the loop on #205 (letting an agent see images referenced from notes) along the way.

## What changes

- **`vault_read_binary` returns a raster image as an `image` block.** It's downscaled on Obsidian's own canvas to fit 1568px on the long edge (Claude resizes to that anyway, so anything larger is wasted bytes), and comes with a small text block giving the path, MIME type, size, and dimensions. A multi-megabyte photo is now a couple of thousand tokens and the model can actually look at it.
- **An SVG is passed through unchanged.** It comes back as its source text in a `resource` block (`image/svg+xml`), byte-for-byte what's in the vault; the canvas never touches it. The model reads the markup directly.
- **Everything else becomes a link or an embedded resource.** With signed URLs on, a non-image file comes back as a `resource_link` to a signed download URL, so the bytes never enter the conversation. With them off, a file under 1 MiB is embedded as a `resource` block; larger ones are refused with a pointer at the REST endpoint. `as: "bytes"` and `as: "link"` override the default.
- **`vault_write_binary` is gone.** Emitting base64 as output tokens is hopeless beyond a few KB, and the agent's host has the file on disk anyway. `vault_get_upload_url` mints a signed, single-use `PUT` URL plus a ready-to-run `curl` command instead; `vault_get_download_url` does the same for `GET`. Both tools only exist while the setting is on.
- **Signed URLs** are a new Advanced setting, on by default, with a configurable lifetime (300s default). The signature is an HMAC-SHA256 over method, normalized vault path, and expiry, under a secret generated per plugin load and kept in memory only, so links die on restart. The host is deliberately _not_ signed so the same link works from whichever hostname the client uses. `PUT` links are consumed on the first 2xx; `GET` links are reusable until expiry and served `inline` (`download=1` for an attachment).
- **`vault_write` and `vault_append` refuse binary targets.** A path whose extension names an image (other than SVG, which is text), audio, video, font, PDF, or archive type, or content with a NUL byte, is refused with a pointer at the upload route. This is the half of the read-as-text-then-write-back bug that #346 didn't cover, and the actual thing that destroyed people's attachments.
- Tool descriptions from #345 are trimmed; README and OpenAPI docs updated.

I also considered and rejected `vault_import(source_path)` / `vault_fetch(url)` tools that would read from outside the vault on the server side; anyone holding the API key (including a prompt-injected agent) could then copy `~/.ssh/*` into the vault, and server-side fetch is SSRF against localhost. The API key's blast radius stays "the vault".

## Things worth a look

- **How the tools learn the request's host.** The v2 MCP SDK gives tool callbacks no view of the transport, so `McpHandler.handleRequest` runs each request inside an `AsyncLocalStorage` store that the signed-URL tools read back to build `scheme://host`. There are HTTP-level unit tests proving the context survives the SDK on both the sessionless and sessionful legs, and the integration run below exercised it against the real plugin, but it's the one piece of this that leans on Node async-context semantics rather than anything explicit.
- **Path normalization.** Signing happens over `normalizeVaultFilePath(path)` and verification re-normalizes the request's decoded path the same way, so `a/../b.png` and `b.png` are one signature and a link can't be minted for one spelling and redeemed for another file. There's a test for the traversal-shaped case on both the REST and MCP sides.
- **Signed GET requests are served `inline`.** API-key requests keep the `attachment` disposition they've always had; only the new signed path changes, since a link exists to be opened.
- **`setSignedUrlsEnabled`** registers/removes the two tools live when the toggle changes, using the existing dynamic-registration path, so a connected client sees a `tools/list_changed` rather than needing a plugin reload.
- **The default is on.** `enableSignedUrls: true` sits in `DEFAULT_SETTINGS`, so an existing `data.json` with no key comes up enabled, and the settings tab stores `false` explicitly when the toggle is turned off.

## Verification

- 578 unit tests pass (`npm test`), including 28 for the signing module, 11 for the scaler, and coverage across `mcpHandler.test.ts` and `requestHandler.test.ts` for every branch above, plus the SVG pass-through and its two fall-through cases.
- The full integration suite ran against a live Obsidian with this branch loaded — first with signed URLs enabled by hand, then after the review round with untouched default settings: 250 passed, 7 skipped (the usual opt-in `active_file`/`open_file` ones), 1 failed. The failure is `cors.test.ts` › "a POST that returns Content-Location exposes it to a browser", which asserts a `Content-Location` header on `POST /vault/<path>/heading/…`. Nothing on that route sets one (only the `/active/` handlers and MOVE/COPY do), and this branch doesn't touch it, so I believe that's pre-existing from #344; worth confirming on `main`.
- The live run is what shows the canvas path works: the 1×1 PNG came back as an `image` block with `width: 1, height: 1`, an upload link accepted exactly one unauthenticated `PUT` and refused the replay, the download link served the bytes back `inline`, and a real `.svg` came back as its own source text.

## Honest gaps

- The scaler is only exercised for real on a 1×1 PNG. A large JPEG/PNG re-encode through `OffscreenCanvas.convertToBlob` is covered by unit tests against a fake canvas, not against Electron's real one.
- `X-Forwarded-Proto` is honoured for the scheme with no trust-proxy check. Since the only party a forged header misleads is the authenticated caller who sent it, I left it simple, but it's a choice.


docs/src/lib/descriptions/mcp.md
```diff
@@ -51,11 +52,17 @@ Requests with an unrecognized `MCP-Protocol-Version` value are rejected with `40
...
 
-Base64 in a tool argument or result passes through the model's context at roughly 0.35-0.45 tokens per byte, so both tools refuse files over 1 MiB. That is a context guard rather than a storage limit: `GET` and `PUT /vault/{filename}` carry raw bytes of any size at no context cost, and are the right way to move a large attachment.
+There is no upload tool that carries bytes through the model. Upload a file with `PUT /vault/{filename}` — with the API key, or with a signed upload URL from `vault_get_upload_url`.
+
+### Signed URLs
+
+Off by default; enabled under Advanced settings, where the lifetime is also set (default 300 seconds). While enabled, `vault_get_download_url` and `vault_get_upload_url` are registered, and `vault_read_binary` links to non-image files instead of embedding them.
```
coddingtonbear: lets make these on by default.

src/imageScaling.ts
```diff
@@ -0,0 +1,127 @@
+/**
```
coddingtonbear, line 1: In the PR description you say that SVG images are also scaled.  Is that really necessary? Being vector graphics, I would assume we could pass them through directly unchanged.

coddingtonbear (changes requested):

docs/src/lib/descriptions/mcp.md
coddingtonbot: Done in f4c9fcf. `enableSignedUrls` now lives in `DEFAULT_SETTINGS` as `true`, so an existing `data.json` without the key comes up with them on, and the toggle stores `false` explicitly when you turn it off (the old `|| undefined` would have read back as the default and made it impossible to disable). README, this page, and the tool-count assertions are updated to match. The live run in the evidence comment below picked this up on its own: the plugin came up with the two link tools registered without my touching your settings.

src/imageScaling.ts
coddingtonbot: You're right that there's nothing to gain from scaling a vector, and it turns out the rasterizing path was never sound anyway: Chromium's `createImageBitmap` refuses an SVG blob unless it has an intrinsic width and height ([issues.chromium.org/40461578](https://issues.chromium.org/issues/40461578)), and it was on the honest-gaps list as never run live. The one thing rasterizing would have bought is the model *seeing* the drawing — an `image` block can only carry PNG/JPEG/GIF/WebP, so an SVG can't go through that path as-is.

So as of f259be7 an SVG is passed through unchanged: `vault_read_binary` returns its source as a `text` resource (`image/svg+xml`), byte-for-byte what's in the vault, and the scaler never sees it. The model reads the markup, which for the diagrams and Excalidraw exports that actually live in vaults is arguably the more useful form anyway. Over the 1 MiB ceiling, or not valid UTF-8, it falls through to the link/bytes path like any other non-image file; `as: "bytes"` and `as: "link"` behave as before.

One consequence I took along: since SVG is text, `vault_write` and `vault_append` no longer refuse `.svg` paths — the binary-extension guard exempts `image/svg+xml`. Say the word if you'd rather keep that guard as it was.

PR #358, opened 2026-09-21
Refuse MCP vault paths that escape the vault root (GHSA-j9qv-qgpv-345g)

coddingtonbot:
Fixes the path traversal reported in GHSA-j9qv-qgpv-345g.

The advisory reported that the MCP `vault_write` tool will write to any path the Obsidian process can reach: a `path` argument of `../../Ausserhalb.md` lands two directories above the vault and comes back with `{"message": "OK"}`. It also noted that every other mutating tool the reporter tried refused the same input, and concluded that `vault_write` was a lone regression in the MCP tool layer.

It turns out that isn't quite right, and the reason it looked that way is worth explaining, because it's what decided where the fix goes.

# Why most of the tools looked safe

Obsidian's Vault API isn't a sandbox, and its two halves disagree about what an out-of-vault path means:

- `getAbstractFileByPath` consults the index of files Obsidian already knows about. `../secrets.md` isn't in that index, so the lookup misses and the caller is told "File not found". That covers `vault_read`, `vault_patch`, and a _trashing_ `vault_delete` — they look validated, but they're only lucky.
- `Vault.create`, `Vault.adapter.writeBinary` and `Vault.adapter.remove` do no lookup at all. They join the path onto the vault directory and hand it to the filesystem, where `..` means what it always means.

So the hole was wherever that second group was reachable, and that's three tools, not one:

| Tool | Reaches | Impact |
| --- | --- | --- |
| `vault_write` | `Vault.create` / `adapter.writeBinary` | arbitrary write (the reported case) |
| `vault_append` | `Vault.create`, when the target doesn't already exist | arbitrary write |
| `vault_delete` with `permanent: true` | `adapter.exists` + `adapter.remove` | arbitrary **delete** |

The permanent-delete case is strictly worse than the reported write, and it's invisible from the outside: the default `permanent: false` takes the trashing branch and answers "File not found", which is almost certainly why the reporter's probe cleared `vault_delete`.

This is distinct from GHSA-62gx-5q78-wrvx (fixed in 4.1.3), which was an Express routing and URL-decoding gap on the REST `/vault/{path}` endpoints. The guard added then still works; the MCP tool layer just never grew an equivalent, apart from the `vault_move`/`vault_copy` destinations, which each hand-rolled their own copy of the check.

# The fix: one rule, three layers

The advisory suggested applying the `vault_move`/`vault_copy` check to `vault_write`, "ideally centralized". Centralizing it at the MCP boundary alone would leave the same class of bug one forgotten tool away, and wouldn't cover the REST handler or the extension API at all. So the rule lives in a new leaf module, `src/vaultPath.ts`, and three layers call it:

- **`VaultOperations` checks every path it's handed.** This is what actually closes the hole. It's the chokepoint that the REST handler, the MCP tools and any plugin holding the extension API all funnel through, so a boundary somebody forgets to guard can't reach the filesystem through it.
- **Each MCP tool checks first**, so a client gets a refusal naming the offending field rather than a bare failure. The two hand-rolled destination checks are now this same call, and their _source_ paths are checked too, which they weren't before.
- **The REST handler's existing checks are the same call**, and the express error handler maps `PathTraversalError` onto the `40021` it already returns, so the `VaultOperations` backstop can't surface as a 500.

The second and third layers are deliberately redundant with the first. The first is the security property; the other two are error quality.

# Two behavior changes beyond the reported bug

- **Backslashes fold to `/` before resolving.** On Windows the adapter treats `..\` as a separator that `posix.resolve` doesn't. Folding only ever makes a path look _more_ contained — a POSIX file genuinely named `a\b.md` gets checked as `a/b.md`, which is still inside — so it can't reject anything that was safe.
- **An absolute path is refused outright** rather than resolved. The old REST check resolved against a synthetic `/vault` root, which meant `/vault/notes/a.md` passed by coincidence of spelling. A vault-relative path never begins with `/`.

# Known limitation: symlinks

The advisory asked for a real-path containment check. This is a textual one. A path that stays inside the vault as written can still point outside it through a symlinked folder, and Obsidian's API doesn't expose a real-path primitive to check that with. I've judged that acceptable, since a symlink inside the vault is something the vault's owner put there deliberately — but it is a real gap and worth knowing about.

# Testing

TDD, per `AGENTS.md`: the failing tests went in first (109 of them), then the fix.

- `src/vaultPath.test.ts` covers the rule itself.
- `src/vaultOperations.test.ts` covers the chokepoint, and asserts that no create, write or remove call is made at all — not just that an error came back.
- `src/mcpHandler.test.ts` covers every path-taking tool, table-driven across several shapes of escaping path, including `vault_read_binary` and the signed-URL tools.
- `src/integration/mcp.test.ts` adds a live probe. It uses a single-level traversal with a distinctive filename on purpose, so that a regression leaves something obvious next to the vault directory rather than clobbering a real file.

823 unit tests pass; `typecheck` and `lint` are clean. The integration suite has now been run against a live Obsidian with this branch built and loaded: 13 suites, 273 passed, 7 skipped (the `active_file_*` and `open_file` tests that opt in by env var). Nothing appeared outside the vault.

# On `normalizeVaultFilePath`

`src/signedUrls.ts` already carried its own containment check inside `normalizeVaultFilePath`, which is why `vault_read_binary` and the two signed-URL tools were never exposed to this. That check is now the shared one, so there's a single definition in the codebase of what escapes the vault — two layers disagreeing about that is what produced the bug in the first place.

It keeps the rest of its own job, though. It's a canonicaliser rather than just a validator, so it still refuses a directory or an empty path, and it still _strips_ a leading slash where everything else refuses one — a signature minted for `a/b.png` has to verify a request for `/a//b.png`, which is a sloppy spelling of the same file rather than an escape from anywhere. `signedUrls.test.ts` pins that on both sides, and I've left it alone.


src/mcpHandler.ts
```diff
@@ -831,59 +389,16 @@ export class McpHandler {
...
-      ),
-    ];
-  }
-
   private registerTools(): void {
     this.tool(
       "vault_list",
       dedent`List files and subdirectories inside a vault directory. Returns an array of names; directory entries end with '/'. Omit path or pass an empty string to list the vault root.`,
-      { path: z.string().optional().describe("Directory path relative to vault root (default: root)") },
+      { path: z.string().optional().describe("Directory path relative to vault root; must not escape the vault root (default: root)") },
```
coddingtonbear: I don't think it's necessary for your to call out this new fact explicitly. That's just the expectation.

src/mcpHandler.ts
```diff
@@ -277,6 +98,13 @@ interface Session {
   toolHandles: Map<string, RegisteredTool>;
 }
 
+// Every tool that takes a vault path says the same thing about it, so a client that
+// hands one a "../" gets told why before the call rather than after.
+const VAULT_PATH_DESCRIPTION =
+  "File path relative to vault root; must not escape the vault root";
+const SOURCE_VAULT_PATH_DESCRIPTION =
+  "Source file path relative to vault root; must not escape the vault root";
```
coddingtonbear: I think your swapping over of the raw strings to this variable is fine, but you do not need to call out that they can't escape the vault root.  That's just a standard expectation.

coddingtonbear (approved):

src/mcpHandler.ts
coddingtonbot: Agreed — dropped in a307497. The constants stay, they just say what they said before; the `vault_move`/`vault_copy` destination descriptions are untouched, since that sentence predates this branch.

src/mcpHandler.ts
coddingtonbot: Agreed — dropped in a307497. The constants stay, they just say what they said before; the `vault_move`/`vault_copy` destination descriptions are untouched, since that sentence predates this branch.

src/vaultOperations.ts
```diff
@@ -155,6 +156,18 @@ export class VaultOperations {
...
+  /** Refuse a client-supplied path that resolves outside the vault.
+   *
+   *  Every caller -- the REST handler, the MCP tools, a plugin holding the
+   *  extension API -- funnels through this class, so the containment check lives
+   *  here as well as at each boundary. The boundaries exist to give a caller a
+   *  well-shaped error (a 400 with errorCode 40021, a legible MCP tool error);
+   *  this exists so that a boundary someone forgets to guard cannot reach the
+   *  filesystem. See ./vaultPath for why Obsidian's own API does not stop it. */
+  private assertContained(filePath: string, label = "Path"): void {
+    assertVaultPathIsContained(filePath, label);
```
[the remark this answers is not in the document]
coddingtonbot: Good catch — fixed in a83ef8a, with coverage in `vaultOperations.test.ts`.

Worth noting it was not exploitable: `readBinaryFileContent` looks the file up with `getAbstractFileByPath` before touching the adapter, so `../x.png` misses the index and throws "File not found" long before `adapter.readBinary` sees it. But that is precisely the reasoning this branch exists to stop relying on — `readFileContent` has the same lookup and was guarded anyway — so you are right that it should not be the one method opting out.

src/mcpHandler.ts
```diff
@@ -277,6 +277,13 @@ interface Session {
...
```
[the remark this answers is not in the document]
coddingtonbot: Declining this one. The repository owner asked for the opposite in review: a path argument staying inside the vault is the standard expectation rather than news, so restating it on every `path` description spends a client tokens to be told what it already assumed. The clause was removed from the shared descriptions in a307497, which means `vault_read_binary` and the signed-URL tools are already consistent with the rest — just in the other direction from the one suggested here.

The `vault_move`/`vault_copy` destination descriptions keep the sentence, as they did before this branch: a destination is the one path a caller composes rather than echoes back.

src/vaultPath.ts
```diff
@@ -0,0 +1,62 @@
...
+ *  root by coincidence of spelling and has nothing to do with where the vault
+ *  actually lives. A vault-relative path never begins with "/".
+ *
+ *  What this does not do is resolve symlinks: a vault-relative path that stays
+ *  inside the vault textually can still point outside it through a symlinked
+ *  folder. Obsidian's API exposes no real-path primitive to check that with, and
+ *  a symlink inside the vault is something the vault's owner put there. */
+export function vaultPathIsContained(candidate: string): boolean {
+  const normalized = candidate.replace(/\\/g, "/");
+  if (normalized.startsWith("/")) return false;
```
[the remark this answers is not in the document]
coddingtonbot: Fixed in ae82954, with regression tests for `C:/outside.md`, `C:\outside.md`, `c:/outside.md` and the drive-relative `C:outside.md`.

One correction to the reasoning, for the record: it was probably not reachable as it stood. Obsidian's adapter joins with `path.join`, which concatenates rather than re-rooting — `path.win32.join("C:\\vault", "C:/outside.md")` gives the nonsense `C:\vault\C:\outside.md` and fails — rather than `path.resolve`, which is what would actually re-root it.

Fixing anyway, because the comment on `vaultPathIsContained` already claimed an absolute path is refused outright, and that was untrue on Windows. The rule should hold on its own terms rather than on an Obsidian internal staying as it is.

No legitimate path is lost: the check is anchored, so `notes/C:x.md` and `CC:notes.md` still pass, and Windows does not permit `:` in a filename anyway.

